PR #2, opened 2019-01-19
Implements ResourceRegistry and OperationProcessor

joelalejandro:
According to what was discussed in the proposed spec, I've implemented `OperationProcessor` and `ResourceRegistry`.

This is how would it look on an API implementation.

### A scaffolding proposal
```
|_ src
|   |_ resources
|   |   |_ user
|   |   |   |_ resource.ts
|   |   |   |_ processor.ts
```

### The `User` resource

```ts
// src/resources/user/resource.ts
import { Resource } from "@ebryn/jsonapi-ts";

export default interface User extends Resource {
  attributes: {
    username: string;
    ethAddress: string;
    nodeAddress: string;
    email: string;
    multisigAddress: string;
    token?: string;
  };
}
```

### The `UserProcessor` operation processor

```ts
// src/resources/user/processor.ts
import {
  AuthenticatedContext,
  Authorize,
  JsonApiDocument,
  OperationProcessor
} from "@ebryn/jsonapi-ts";

import User from "./resource";

export default class UserProcessor extends OperationProcessor<User> {
  @Authorize()
  async get(id: string, ctx: AuthenticatedContext) {
    const user = ctx.user as User;

    return {
      data: [user]
    } as JsonApiDocument<User>;
  }

  async add(user: User) {
    // Do some db ops.
    const newUser = await db.users.create(user);

    return {
      data: newUser
    } as JsonApiDocument<User>;
  }
}
```

### Registering the Processor

This is an example of a fully-fledged API with our newest additions:

```ts
import { ResourceRegistry } from "@ebryn/jsonapi-ts";
import cors from "@koa/cors";
import Koa from "koa";
import bodyParser from "koa-body";
import Router from "koa-router";
import authenticate from "./middlewares/authenticate";
import UserProcessor from "./resources/user/processor";

const api = new Koa();

const router = new Router({ prefix: "/api" });

// We create a resource registry.
const processors = new ResourceRegistry(router);

api
  .use(bodyParser({ json: true }))
  .use(cors({ keepHeadersOnError: false }))
  .use(authenticate())
  .use(
    processors
      .register("user", UserProcessor) // We add our resource.
      .getEndpoints() // This will soon become JSONAPIKoa()
  );
```


Exelord:
Hi Joel :)
I'm not sure if the spec is fixed already, but I might have some idea to bring more flexibility in creating api servers. The current concept is very similar to what I was planning to build as a ruby gem some time ago, (what is very exciting) but it might have a limitation, I see now.

On most projects I worked on it was very popular to have multiple API servers (aka API versions) for eg. private and a public purpose. Thanks to this approach they could control what kind of information are for 3rd party usage but also develop new API versions without instantly dropping current specification. They also will be able to switch endpoint by endpoint rising the API deprecations being still be backward compatible.

I'm not sure if you guys are fan of this practice but it came out it's quite popular and we may want to cover it, and there is more than this one specific advantage.

My proposal is to add one more level of abstraction (if it's already covered please let me know 😅), called `server` || `version` (how you like) on top of resources. Let me explain it's purpose starting from files structure:
```
src
└── servers
    ├── index.ts (aka default server [in case users don't want to use other namespaces])
    ├── resources
    │   └── user
    │       ├── resource.ts
    │       └── processor.ts
    ├── v1 (user custom server's namespace name)
    │   ├── index.ts (server)
    │   └── resources
    │       └── company
    │           ├── resource.ts
    │           └── processor.ts
    └── private (user custom server's namespace name)
        ├── index.ts (server)
        └── resources
            └── account
                ├── resource.ts
                └── processor.ts
```

The server file (`index.ts`) can be responsible for registering all it's resources and setting the routes. Basically what you mentioned in `Registering the Processor` but without setting router. Then users can choose and import fully isolated servers and set them on the router under specific namespace. Thanks to it you will be able to develop mutiple API versions in one project and still have the mentioned fexibility.

Please let me know your thoghts :) Of course there is still bunch of stuff to dicuss around this thing, but I just wanted to mention this idea and solve one of the use case.

PR #9, opened 2019-01-23
Knex processor

Exelord:


src/knex-processor.ts
```diff
@@ -0,0 +1,54 @@
+import * as Knex from "knex";
+
+import OperationProcessor from "./operation-processor";
+import { Resource } from "./types";
+
+export default class KnexProcessor<
+  ResourceT extends Resource
+> extends OperationProcessor<ResourceT> {
+  private knex: any;
```
joelalejandro: @Exelord can we be explicit about the type for this property? It should be:

```suggestion
  private knex: Knex;
```

src/knex-processor.ts
```diff
@@ -0,0 +1,54 @@
+import * as Knex from "knex";
```
joelalejandro: @Exelord A regular `import` (no wildcard) should work:

```suggestion
import Knex from "knex";
```

If it doesn't, can you verify that `allowSyntheticImports` is set to `true` on the `tsconfig.json` file?

joelalejandro:
Seems that this PR is no longer necessary due to #10?

src/knex-processor.ts
Exelord: It is present by the `*` is still required to have it working :/

Exelord:
Yes, you right!

PR #10, opened 2019-01-23
Application

Exelord:
Dependency on: #9


tests/dummy/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import Koa from "koa";
+
+import { Application, JsonApiKoa, KnexProcessor } from "./jsonapi-ts";
```
joelalejandro: @Exelord Maybe we need to comment the `JsonApiKoa` import for now?

src/types.ts
```diff
@@ -97,8 +97,6 @@ export type JsonApiError<ErrorCodeT = ErrorCode> = {
   detail: string;
 };
 
-export type Application = {};
-
 export type Links = {};
 
 export type Meta = {};
```
joelalejandro: @Exelord We can specify a stricter typing here:

```suggestion
export type Meta = { [key: string]: string | number | boolean | Meta };
```

src/json-api-koa.ts
```diff
@@ -0,0 +1,50 @@
...
+export default function jsonApiKoa(app: Application) {
+  const URL_REGEX = new RegExp(app.types.map(t => t.name).join("|"), "i");
+
+  return async (ctx: Context, next: Function) => {
+    if (ctx.url.match(URL_REGEX) && app.types.length) {
+      const ops: Operation[] = convertHttpRequestToOperations(ctx);
+      const result: OperationResponse[] = await app.executeOperations(ops);
+      ctx.body = convertOperationsResponsesToHttpResponse(ctx, result);
+    } else if (ctx.url.match("/bulk")) {
+      return {
```
joelalejandro: @Exelord this should be `ctx.body`

src/json-api-koa.ts
```diff
@@ -0,0 +1,50 @@
...
+        operations: await app.executeOperations(
+          ctx.request.body.operations || []
+        )
+      };
+    }
+
+    await next();
```
joelalejandro: @Exelord This should be `return next()`

src/json-api-koa.ts
```diff
@@ -0,0 +1,50 @@
...
+  };
+}
+
+function convertHttpRequestToOperations(ctx: Context) {
+  if (ctx.method === "GET") {
+    const [, type, id] = ctx.url.split("/");
```
joelalejandro: @Exelord Let's use [route parameter names](https://github.com/alexmingoia/koa-router#routerurlname-params-options--string--error) for this. That's why the `koa-router` package is included as part of the framework.

You can do:

```ts
const { type, id } = ctx.params;
```

Instead of declaring `ctx: Context`, you'll need to use `ctx: RouterContext`, which is imported from `koa-router`.

src/knex-processor.ts
```diff
@@ -22,7 +22,7 @@ export default class KnexProcessor<
       .select();
   }
 
-  async remove(data: Resource): Promise<boolean> {
+  async remove(data: Resource): Promise<null> {
```
joelalejandro: @Exelord I think it's more semantically correct to return a `boolean` here. However, if you mean to expect no return value, it's probably better to use `Promise<void>`.

src/types.ts
```diff
@@ -123,3 +115,7 @@ export type Operation = {
   links: Links;
   meta: Meta;
 };
+
+export type OperationResponse = {
+  data: Resource | Resource[] | null;
```
joelalejandro: @Exelord I'm uncertain that we want to allow `null` as a value. I'd rather mark it as `undefined` with `data?`.

src/json-api-koa.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    await next();
+  };
+}
+
+function convertHttpRequestToOperations(ctx: Context) {
+  const [, type, id] = ctx.url.split("/");
+
+  if (ctx.method === "GET") {
```
joelalejandro: @Exelord We can apply some DRY improvements to reduce this function a bit.

So, any HTTP method correlates to a JSONAPI operation:

```ts
const methodsToOperations = {
  GET: "get",
  POST: "add",
  DELETE: "remove",
  PATCH: "update",
  PUT: "update"
};
```

Every operation follows pretty much the same structure, except for `data`, which is present on the write operations, so we can do something like this:

```ts
const op = methodsToOperations[ctx.method];

return [
  {
    op,
    ref: { id, type },
    // Inject `data` only for write operations.
    ...(op === "create" || op === "update" ? { data: ctx.request.body.data } : {})
  } as Operation
];
```

src/json-api-koa.ts
```diff
@@ -0,0 +1,90 @@
...
+      } as Operation
+    ];
+  }
+}
+
+function convertOperationsResponsesToHttpResponse(
+  ctx: Context,
+  operations: OperationResponse[]
+) {
+  const responseMethods = ["GET", "POST", "PATCH", "PUT"];
```
joelalejandro: @ebryn Should we worry about CORS here?

src/knex-processor.ts
```diff
@@ -1,47 +1,75 @@
...
-  private knex: any;
+  private knex: Knex;
 
   constructor(public knexOptions = {}) {
     super();
 
     this.knex = Knex(knexOptions);
   }
 
+  private convertToResources(type: string, records: any[]) {
```
joelalejandro: @Exelord Let's define a `Record` type with an `id` property.

src/application.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.types = settings.types || [];
+    this.processors = settings.processors || [];
+    this.defaultProcessor = settings.defaultProcessor;
+
+    this.defaultProcessor.app = this;
+    this.processors.forEach(processor => (processor.app = this));
+  }
+
+  async executeOperations(ops: Operation[]): Promise<OperationResponse[]> {
+    return await Promise.all(
```
ebryn: should we have `createTransaction` here and have it call `Promise.all` for now?

src/application.ts
```diff
@@ -0,0 +1,42 @@
...
+      ops.map(async op => {
+        const processor = this.processorFor(op);
+        const result = await processor.execute(op);
+        return this.buildOperationResponse(result);
+      })
+    );
+  }
+
+  private processorFor(op: Operation): OperationProcessor {
+    return this.defaultProcessor;
```
ebryn: `this.types.find(t => t.shouldHandle(op))` ?

tests/dummy/src/index.ts
```diff
@@ -0,0 +1,25 @@
+import * as Koa from "koa";
+import * as koaBodyParser from "koa-bodyparser";
+
+import { Application, jsonApiKoa, KnexProcessor } from "./jsonapi-ts";
+import User from "./resources/user/resource";
+
+const app = new Application({
+  types: [User],
+  processors: [],
```
ebryn: perhaps we should combine `types` and `processors` to be an object literal: 
```
processors: {
  User: KnexProcessor
  default: SomeOtherProcessor
}```

File: tests/dummy/src/index.ts [hunk 14]
```diff
@@ -0,0 +1,25 @@
...
+    connection: {
+      filename: "./tests/dummy/db/dev.sqlite3"
+    },
+    useNullAsDefault: true
+  })
+});
+
+const koa = new Koa();
+
+koa.use(koaBodyParser());
```
ebryn: lets try to remove this and hide it in `jsonApiKoa`

src/json-api-koa.ts
ebryn: No

src/knex-processor.ts
```diff
@@ -1,47 +1,75 @@
...
+    const records = await this.knex(tableName)
       .where(this.filtersToKnex(filters))
       .select();
+
+    return this.convertToResources(type, records);
   }
 
-  async remove(data: Resource): Promise<boolean> {
-    const tableName = this.typeToTableName(data.type);
+  async remove(op: Operation): Promise<null> {
```
ebryn: `Promise<void>`

src/json-api-koa.ts
Exelord: hmm, according to the doc it's ok. Am I missing sth?
```
app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  ctx.set('X-Response-Time', `${ms}ms`);
});
```

src/types.ts
Exelord: We can cover that later when we gonna work on Meta object :) Right now it's just to not forget about it.

src/json-api-koa.ts
Exelord: `ctx.body` is an alias for `ctx.reponse.body`. Here we want to use the `ctx.request.body` - the incoming data

src/json-api-koa.ts
Exelord: I suggested that also to Erik but we decided to move on with the simplest version possible, to have the POC ready asap.
and `ctx.url` is a relative path to host so it's not a case here, fortunately :) But yes I agree, routing will be better here.

src/knex-processor.ts
Exelord: I was thinking about that, but as `remove` is a promise there is no need for any response. The promise will resolve or not. There won't be any chance to get other response than `true` (1) so it doesn't make sense to use it. Would you agree?

src/types.ts
Exelord: According to JSONAPI doc, there are cases when we need to return `data: null` so it might be one of those.

src/json-api-koa.ts
Exelord: Oh yeah for sure! It's just the simplest way possible to make it works. I think we gonna start improving the implementation later, but it depends what Erik will say.

src/application.ts
Exelord: So types should be aware of processors?

src/json-api-koa.ts
joelalejandro: Probably won't make a difference, nvm :)

src/application.ts
ebryn: oops that should have been `this.processors.find`

src/knex-processor.ts
```diff
@@ -0,0 +1,89 @@
...
+import OperationProcessor from "./operation-processor";
+import Resource from "./resource";
+import { KnexRecord, Operation, ResourceConstructor } from "./types";
+
+export default class KnexProcessor<
+  ResourceT extends ResourceConstructor
+> extends OperationProcessor<ResourceT> {
+  private knex: Knex;
+
+  constructor(public knexOptions = {}) {
```
joelalejandro: @Exelord This should be of type `knex.Config` for better intellisense

src/types.ts
```diff
@@ -63,35 +39,37 @@ export type ResourceRelationship = {
...
 
-export type JsonApiErrorDocument<ErrorCodeT = ErrorCode> = {
-  errors?: JsonApiError<ErrorCodeT>[];
-  meta?: Metadata;
-};
-
-export type JsonApiRequest = {
-  data?: Resource;
-  meta?: Metadata;
+export type JsonApiErrorsDocument = {
```
joelalejandro: @Exelord this seems a bit counterintuitive in terms of the JSONAPI spec, which says that a document _may_ have one of four top-level elements: `data`, `meta`, `included`, `errors`. Either we bring the `errors[]` collection back to `JsonApiDocument` or we make it a mandatory field in `JsonApiErrorsDocument` (`errors` being optional in this type makes little sense).

src/types.ts
```diff
@@ -63,35 +39,37 @@ export type ResourceRelationship = {
...
 };
 
-export type JsonApiError<ErrorCodeT = ErrorCode> = {
-  status: HttpStatusCode;
-  code: ErrorCodeT;
-  title: string;
-  detail: string;
+export type JsonApiError = {
+  id?: string;
+  status?: HttpStatusCode;
```
joelalejandro: @Exelord Both `status` and `code` should be mandatory.

src/types.ts
Exelord: Do not forget these rules:
> The data member MUST NOT coexist with either the errors or operations members in the same document.

> The operations member MUST NOT coexist with the errors member in the same document.

So basically there is no chance data can coexist with errors.

Yes I agree, JsonApiErrorsDocument should require errors :)

src/types.ts
Exelord: > An error object MAY have the following members:
...
status: the HTTP status code applicable to this problem, expressed as a string value.
code: an application-specific error code, expressed as a string value.
...

Why should we require them?

joelalejandro: Beacuse otherwise there's no point on sending an error response if you don't have at very least a status and code.

PR #30, opened 2019-02-19
[poc] Manual support for including resources

joelalejandro:


src/application.ts
```diff
@@ -30,7 +30,7 @@ export default class Application {
       ops.map(async op => {
         const processor = this.processorFor(op);
         const result = await processor.execute(op);
-        return this.buildOperationResponse(result);
+        return this.buildOperationResponse(result, processor.flushIncludes());
```
ebryn: any particular reason we need to "flush" the includes?

joelalejandro: Just to make sure the included resources don't survive the current operation. I'm probably being over-defensive.

PR #31, opened 2019-02-20
filters

aklkv:
fixes: #17 

this is very much work in progress, feedback in encouraged

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@koa/cors": "^2.2.3",
     "camelize": "^1.0.0",
     "dasherize": "^2.0.0",
+    "ember-cli-string-utils": "^1.1.0",
     "escape-string-regexp": "^1.0.5",
     "jsonwebtoken": "^8.4.0",
     "knex": "~0.16.3",
--- a/src/processors/knex-processor.ts
+++ b/src/processors/knex-processor.ts
@@ -6,6 +6,20 @@ import { KnexRecord, Operation, ResourceConstructor } from "../types";
 
 import OperationProcessor from "./operation-processor";
 
+const operators = {
+  eq: '=',
+  ne: '!=',
+  lt: '<',
+  gt: '>',
+  le: '<=',
+  ge: '>=',
+  like: 'like',
+  in: 'in',
+};
+
+const getOperator = (paramValue: any) =>
+  operators[Object.keys(operators).find(operator => paramValue.indexOf(`${operator}:`) === 0)];
+
 export default class KnexProcessor<
   ResourceT = Resource
 > extends OperationProcessor<ResourceT> {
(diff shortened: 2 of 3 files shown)


src/processors/knex-processor.ts
```diff
@@ -6,6 +6,20 @@ import { KnexRecord, Operation, ResourceConstructor } from "../types";
...
 
+const operators = {
+  eq: '=',
+  ne: '!=',
+  lt: '<',
+  gt: '>',
+  le: '<=',
+  ge: '>=',
+  like: 'like',
+  in: 'in',
```
joelalejandro: @aklkv Would be nice to have a `NOT IN` operator.

aklkv: who would it look like in query params and knex?

something like:
```
whereNot((builder) => { 
   builder. whereIn([]);
} 
```

not sure if this would work tho

joelalejandro: I think the query parameter would be `nin` (following `ne` as non-equal), and the SQL would be `not in`.

aklkv: ok I can do that

package.json
```diff
@@ -23,6 +23,7 @@
     "@koa/cors": "^2.2.3",
     "camelize": "^1.0.0",
```
joelalejandro: @aklkv If we add `ember-cli-string-utils`, both `camelize` and `dasherize` should be removed.

src/processors/knex-processor.ts
```diff
@@ -1,11 +1,27 @@
...
+  nin: 'not in'
+};
+
+const getOperator = (paramValue: any) =>
```
joelalejandro: @aklkv This should be typed as `string`.

src/processors/knex-processor.ts
```diff
@@ -1,11 +1,27 @@
...
+  operators[Object.keys(operators).find(operator => paramValue.indexOf(`${operator}:`) === 0)];
```
joelalejandro: @aklkv We can use `.startsWith()` instead of `.indexOf()` - feels more readable.

src/processors/knex-processor.ts
```diff
@@ -82,11 +98,38 @@ export default class KnexProcessor<
...
     Object.keys(filters).forEach(
-      key => filters[key] === undefined && delete filters[key]
+      key => filters[key] === undefined && delete filters[key],
     );
 
-    return filters;
+    Object.keys(filters).forEach((key) => {
+
+      let value = filters[key];
+      let operator = getOperator(filters[key]);
```
joelalejandro: @aklkv The following `if` can be removed in favor of:

```ts
const operator = getOperator(filters[key]) || "=";
```

src/processors/knex-processor.ts
```diff
@@ -82,11 +98,38 @@ export default class KnexProcessor<
...
+
+      if (!operator) {
+        operator = '=';
+      }
+
+      if (value.substring(value.indexOf(':') + 1)) {
+        value = value.substring(value.indexOf(':') + 1)
+      }
+
+      value = value !== 'null' ? value : 0;
```
joelalejandro: @aklkv Why are we converting `null` to `0`?

aklkv: it come as a string 'null'

package.json
aklkv: I think this is addressed in #14 I don't mind sticking with just camelize package tho

PR #39, opened 2019-02-26
adding ability to select fields, on single resource only for now

aklkv:
fixes: #25

### DIFF
--- a/src/processors/knex-processor.ts
+++ b/src/processors/knex-processor.ts
@@ -61,10 +61,10 @@ export default class KnexProcessor<
     const { id, type } = ref;
     const tableName = this.typeToTableName(type);
     const filters = params ? { id, ...(params.filter || {}) } : { id };
-    const resource = Object.create(this.resourceFor(type));
+    const resource = this.resourceFor(type);
     const fields = params ? { ...params.fields } : {};
     const attributes = getAttributes(
-      Object.keys(resource.__proto__.attributes),
+      Object.keys(resource.prototype.attributes),
       fields,
       type
     );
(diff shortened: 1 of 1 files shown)


src/processors/knex-processor.ts
```diff
@@ -51,9 +61,13 @@ export default class KnexProcessor<
     const { id, type } = ref;
     const tableName = this.typeToTableName(type);
     const filters = params ? { id, ...(params.filter || {}) } : { id };
-    const Resource = Object.create(this.resourceFor(type));
+    const resource = Object.create(this.resourceFor(type));
```
ebryn: why are we using Object.create?

src/processors/knex-processor.ts
```diff
@@ -42,15 +57,21 @@ export default class KnexProcessor<
...
+    const { id, type } = ref;
     const tableName = this.typeToTableName(type);
-    const filters = op.params ? { id, ...(op.params.filter || {}) } : { id };
-    const Resource = Object.create(this.resourceFor(type));
-    const attributes = Object.keys(Resource.__proto__.attributes);
+    const filters = params ? { id, ...(params.filter || {}) } : { id };
+    const resource = Object.create(this.resourceFor(type));
+    const fields = params ? { ...params.fields } : {};
+    const attributes = getAttributes(
+      Object.keys(resource.__proto__.attributes),
```
ebryn: it seems like if we didn't use `Object.create` we could just be doing something like: 
```
const Resource = this.resourceFor(type);
// ...
Object.keys(Resource.prototype.attributes)```

aklkv: Resource is a subclass apparently, so the only why to access prototype is with `__proto__`

PR #42, opened 2019-03-04
Async Relationships and Computed Property Post Processing

rtablada:


rtablada:
I need to remove the .vscode launch since that's a separate PR and @Exelord is working on that.

src/processors/knex-processor.ts
```diff
@@ -73,7 +70,7 @@ export default class KnexProcessor<
       .select(...attributes, "id")
       .modify(queryBuilder => this.optionsBuilder(queryBuilder, op));
 
-    return this.convertToResources(type, records);
+    return records;
```
rtablada: Convert to resources is now a concern of the base class operation processor and is applied to all operations via `execute`.

src/processors/operation-processor.ts
```diff
@@ -11,9 +29,58 @@ export default class OperationProcessor<ResourceT = Resource> {
     return this.resourceClass && op.ref.type === this.resourceClass.name;
   }
 
-  execute(op: Operation): Promise<ResourceT | ResourceT[] | void> {
+  async execute(op: Operation): Promise<ResourceT | ResourceT[] | void> {
     const action: string = op.op;
-    return this[action] && this[action].call(this, op);
+    let result = this[action] && await this[action].call(this, op);
+
+    return this.convertToResources(op.ref.type, result);
```
rtablada: Here's where we `convertToResources`

tests/dummy/src/resources/user/processor.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      return r
+    });
+
+    return records;
+  }
+
+  attributes = {
+    async friends(user: User) {
```
rtablada: This shows an async computed property. Some examples I've seen in the wild is getting a sum value from other tables, getting values from Redis, etc.

tests/dummy/src/resources/user/processor.ts
```diff
@@ -0,0 +1,75 @@
...
+      return 3;
+    }
+  }
+
+  relationships = {
+    async posts(this: UserProcessor, user: any) {
+      if (user.posts !== undefined) {
+        console.log('Loading Already Done!');
+        return user.posts;
+      }
```
rtablada: This is a double check to make sure the relationship hasn't already been loaded.

If it's eager loaded already (like it is here) then this becomes a simple return. Else we query out to the db for each user.

tests/dummy/src/resources/user/processor.ts
```diff
@@ -0,0 +1,75 @@
...
+    const filters = params ? { id, ...(params.filter || {}) } : { id };
+    const resource = Object.create(this.resourceFor(type));
+    const fields = params ? { ...params.fields } : {};
+    const attributes = this.getColumns(
+      Object.keys(resource.__proto__.attributes),
+      fields,
+      type
+    );
+
+    let records: KnexRecord[] = await this.knex(tableName)
```
rtablada: This isn't a good loading for actual `hasMany` but I just wanted to get some basic data flowing.

Replace this with better SQL and linking of relations.

Also we should only add joins if the relation is in `include`

src/processors/operation-processor.ts
```diff
@@ -11,9 +29,61 @@ export default class OperationProcessor<ResourceT = Resource> {
...
+  async convertToResources(type: string, records: HasId[] | HasId) {
+    if (Array.isArray(records)) {
+      return Promise.all(records.map(record => {
+        return this.convertToResources(type, record);
+      }));
+    }
+
+    const record = {...records};
+    const id = record.id;
+    const [computedAttributes, relationships] = await Promise.all([
```
rtablada: Here's where we calculate the computed properties and relationships for each resource.

src/processors/operation-processor.ts
```diff
@@ -11,9 +29,61 @@ export default class OperationProcessor<ResourceT = Resource> {
...
+      fields[pluralize(type)].includes(attribute)
+    );
+  }
+
+  get computedPropertyNames() : string[] {
+    return Object.keys(this.attributes);
+  }
+
+  async getComputedProperties(record: HasId) {
+    return promiseHashMap(this.attributes, (key) => this.attributes[key].call(this, record));
```
rtablada: Here we should do more filtering so we only load CPs that have been requested.

So if we request`/users?fields=user.email` we don't need to run the CP for `user.friends`.

src/processors/operation-processor.ts
```diff
@@ -11,9 +29,61 @@ export default class OperationProcessor<ResourceT = Resource> {
...
+  }
+  async getRelationships(record: HasId) {
+    return promiseHashMap(this.relationships, (key) => this.relationships[key].call(this, record));
```
rtablada: This needs to be filtered so only included relations are run.

src/processors/operation-processor.ts
```diff
@@ -1,7 +1,25 @@
 import Application from "../application";
 import Resource from "../resource";
 import { Operation, ResourceConstructor } from "../types";
-import { classify, singularize } from "../utils/string";
+import { classify, singularize, pluralize } from "../utils/string";
+
+export interface HasId {
+  id: any;
```
Exelord: any? 🤔

src/processors/operation-processor.ts
```diff
@@ -22,19 +92,23 @@ export default class OperationProcessor<ResourceT = Resource> {
...
   }
 
-  async add(op: Operation): Promise<ResourceT> {
+  async add(op: Operation): Promise<HasId> {
     return Promise.reject();
   }
+
+  attributes = {}
+
+  relationships = {}
```
Exelord: Do we want to have relationships hash specific for operation in a singleton processor?

tests/dummy/src/resources/user/processor.ts
Exelord: The join will scope the main results, which is not a case I think so. We want to return all the Users and includes are the extend to the query, not the limitation.

tests/dummy/src/resources/user/processor.ts
Exelord: Is there any reason why we dont wont to keep that in the resource itself?

rtablada: Because we don't want resources to know about databases, connections, etc.

src/processors/operation-processor.ts
rtablada: I'm working on the `get` use case first. I think it can go forward from there.

tests/dummy/src/resources/user/processor.ts
rtablada: Ideally unless you need something custom you won't need to declare your own relationships.

tests/dummy/src/resources/user/processor.ts
rtablada: For the joins we need to do something more like https://github.com/bookshelf/bookshelf/blob/fadb11cc7a9d32fb8d72151b2a23e990e6ba95d2/lib/eager.js

That's the source code of when you want to eager load related data with bookshelf (an ORM built on top of KNEX).

Specifically how it works with `withRelated` and maps more queries.

src/processors/operation-processor.ts
rtablada: This probably needs to get the current operation to be able to tell which includes are called?

rtablada: ```this.resourceClass.schema.relationships.filter(relationName => op.ref.includes.contains(relationName))```

PR #43, opened 2019-03-04
Feature includes

Exelord:


src/application.ts
```diff
@@ -28,9 +36,11 @@ export default class Application {
   async executeOperations(ops: Operation[]): Promise<OperationResponse[]> {
     return await this.createTransaction(
       ops.map(async op => {
-        const processor = this.processorFor(op);
-        const result = await processor.execute(op);
-        return this.buildOperationResponse(result);
+        const processor = this.processorFor(op.ref.type);
+        const data = await processor.execute(op);
+        const included = await this.resolveIncludes(op, data);
+
```
ebryn: here is probably where you want to now mutate the resource objects in `data` to "inject" the `relationships` object that provides the links to the resources now in `included`

tests/dummy/src/resources/user.ts
```diff
@@ -0,0 +1,7 @@
+import { Resource } from "../jsonapi-ts";
+
+export default class User extends Resource {
+  static attributes = {
```
rtablada: How are we declaring that a User has many posts or whatever relationship exists between these two resources?

Exelord: We don't have resource relationships yet :/

rtablada: Even if it's something temporary I think we should look at how we declare relations before we can tackle includes.

ebryn: Can we not have these kinds of discussions on GH? That is what Slack is for FFS

PR #44, opened 2019-03-05
chore: declare node version compatibility

rtablada:


joelalejandro:
I don't agree with this restriction. AWS Lambda services support up to Node 8, and we're hardly using Node 10-specific stuff. We should be able to support a lower version.

PR #48, opened 2019-03-26
Use processor shouldHandle

Exelord:
We will use processor#shouldHandle method to determine if an operation can be handled or not.

### Example:
If there will be a processor called `customProcessor` which overrides `shouldHandle`:
```js
  async shouldHandle(op: Operation): Promise<boolean> {
    return op.ref.type === "customResource"
  }
```

That would mean that every operation with type `customResource` will go through this processor.

If the type will be eg: `user` and there will be no any extra processor,  the operation will be handled by `dafaultProcessor` if there is a resource with type `user`. Otherwise, we will return 404 as this operation is not supported by our server.

### DIFF
--- a/src/processors/operation-processor.ts
+++ b/src/processors/operation-processor.ts
@@ -7,7 +7,7 @@ export default class OperationProcessor<ResourceT = Resource> {
   public app: Application;
   public resourceClass?: ResourceConstructor;
 
-  shouldHandle(op: Operation) {
+  async shouldHandle(op: Operation): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


src/processors/operation-processor.ts
```diff
@@ -7,7 +7,7 @@ export default class OperationProcessor<ResourceT = Resource> {
   public app: Application;
   public resourceClass?: ResourceConstructor;
 
-  shouldHandle(op: Operation) {
+  async shouldHandle(op: Operation): Promise<boolean> {
```
joelalejandro: This seems unnecessarily async.

PR #51, opened 2019-03-28
Adds user attribute check for @Authorize

joelalejandro:
This allows to add conditions to the `@Authorize` decorator by checking attribute values on the User model.

**Example usage**
```ts
import {
  KnexProcessor,
  Authorize,
  IfUser
} from "jsonapi-ts";

export default class BookProcessor extends KnexProcessor {
  @Authorize(IfUser("role", "admin"))
  public async add(op: Operation): Promise<any> {
    return super.add(op);
  }
}
```


ebryn:
😍

PR #63, opened 2019-04-03
Use processor instances instead classes

Exelord:


src/processors/operation-processor.ts
```diff
@@ -3,13 +3,14 @@ import Resource from "../resource";
 import { Operation, ResourceConstructor } from "../types";
 
 export default class OperationProcessor<ResourceT = Resource> {
-  public app: Application;
```
rtablada: Why did you remove Application?

src/processors/operation-processor.ts
```diff
@@ -3,13 +3,14 @@ import Resource from "../resource";
...
-  public resourceClass?: ResourceConstructor;
+  static resourceClass?: ResourceConstructor;
 
-  async shouldHandle(op: Operation): Promise<boolean> {
+  static async shouldHandle(op: Operation): Promise<boolean> {
     return this.resourceClass && op.ref.type === this.resourceClass.type;
   }
 
+  constructor(protected app: Application) {}
```
rtablada: Should this constructor set this.app?

src/processors/operation-processor.ts
Exelord: Move to the constructor. Previously we iterating through all processors instances and injecting the app directly on the object like` processor.app = this` in application constructor. Right now we dont have to do it as we have an access to the processor initializer

src/processors/operation-processor.ts
Exelord: >Moved to the constructor. Previously we iterating through all processors instances and injecting the app directly on the object like processor.app = this in application constructor. Right now we dont have to do it as we have an access to the processor initializer

src/processors/operation-processor.ts
rtablada: Don't declare properties in the constructor, don't get clever. Use `protected` in the class definition and then set the value in the constructor.

Exelord: Ok, Im new to typescript and thought that this is the way now as I saw that in TS doc

PR #73, opened 2019-04-10
Adds support for generic loading of relationships via includes

joelalejandro:
**Without relationships**
image

**belongsTo**
image

**hasMany**
image


src/processors/knex-processor.ts
```diff
@@ -198,4 +198,45 @@ export default class KnexProcessor<
...
+      return this.fetchRelationship(relationshipSchema, record);
+    });
+  }
+
+  async fetchRelationship(
+    relationship: ResourceSchemaRelationship,
+    record: HasId
+  ) {
+    if (relationship.belongsTo) {
+      const sourceTable = this.typeToTableName(
```
rtablada: Only nitpick is here let's do `this.resourceFor(relationship.type().type).getQuery() so we don't lose the right info

src/processors/knex-processor.ts
```diff
@@ -198,4 +198,45 @@ export default class KnexProcessor<
...
+        relationship.foreignKeyName || relationship.type().type
+      );
+      return this.knex(sourceTable)
+        .where("id", record.id)
```
rtablada: I think this should be:

```js
query.join(this.tableName, `${this.tableName}.${foreignKeyName}`, '=', `${relationshipTable}.id`)
```

PR #82, opened 2019-04-18
fix relationships

### LINKED ISSUE #81: Included relationships aren't valid JSONAPI
Postman 2019-04-17 21-44-01

spersico:
closes #81


src/application.ts
```diff
@@ -9,6 +9,7 @@ import {
 } from "./types";
 import pick from "./utils/pick";
 import unpick from "./utils/unpick";
+import { finished } from "stream";
```
joelalejandro: @spersico This seems to be an unexpected import? I don't see it being used anywhere.

src/application.ts
```diff
@@ -89,14 +90,17 @@ export default class Application {
   async buildOperationResponse(
     data: Resource | Resource[] | void
   ): Promise<OperationResponse> {
-    const included = flatten(await this.extractIncludedResources(data)).filter(
-      Boolean
-    );
+    const included = flatten(await this.extractIncludedResources(data))
+    const uniqueIncluded = Array.from(new Set(included
```
joelalejandro: @spersico This can be written in spread notation:

```ts
const uniqueIncluded = [...new Set(included)].map(...).map(...)
```

src/application.ts
```diff
@@ -89,14 +90,17 @@ export default class Application {
...
+      .map((item: Resource) => `${item.type}_${item.id}`)))
```
joelalejandro: @spersico Shouldn't this be `foreignKeyName`-aware?

src/application.ts
```diff
@@ -150,20 +156,19 @@ export default class Application {
...
     Object.keys(data.relationships).forEach(relationshipName => {
       if (Array.isArray(data.relationships[relationshipName])) {
-        data.relationships[relationshipName] = (data.relationships[
-          relationshipName
-        ] as any).map(rel => {
-          rel["type"] = schemaRelationships[relationshipName].type().type;
-          return rel;
-        });
+        data.relationships[relationshipName] =
+          (data.relationships[relationshipName] as any).map(rel => {
```
joelalejandro: @spersico Can you add a `// TODO` here so we can take care of this `any` later? (It's probably mine, but just so I don't forget about it.)

src/application.ts
```diff
@@ -200,5 +205,10 @@ export default class Application {
         });
       })
     );
+    const validIncluded = flatten(rawInclude.filter(Boolean));
```
joelalejandro: @spersico This seems a repetition from `Application#buildOperationResponse`'s first three lines. Should we extract into a single, shared function?

src/application.ts
spersico: will do

src/application.ts
spersico: noup, I'm a dumbass, I just forgot to delete the other one

src/application.ts
spersico: @joelalejandro what do you mean by foreignKeyName?
I check only for repeated resources, I assume that the id and type are unique for each resource, and that's what I use, unless there's something that I'm not seeing

src/application.ts
spersico: done

PR #83, opened 2019-04-19
Fix custom named relationships, Fix Has Many includes

### LINKED ISSUE #79: Belongs to relationship keys aren't fetched via their specified `foreignKeyName` fields
Instead they are being computed assuming `Id`-suffixes here: https://github.com/ebryn/jsonapi-ts/blob/dfc30e3922ae9808eecee511667bcdc2ab61d82c/src/processors/knex-processor.ts#L66

spersico:
closes #79  and fixes hasMany include (for example GET /user?include=articles)


src/processors/knex-processor.ts
```diff
@@ -63,8 +63,7 @@ const getColumns = (resourceClass: typeof Resource, fields = {}): string[] => {
 
   const relationshipsKeys = Object.entries(relationships)
     .filter(([key, value]) => value.belongsTo)
-    .map(([key]) => `${key}Id`);
-
+    .map(([key, value]) => value.foreignKeyName || `${key}Id`);
```
ebryn: we should have a hook for this instead of hardcoding `${key}Id`

PR #88, opened 2019-04-22
Fix columns named "type"

### LINKED ISSUE #87: Columns named "type" are ignored in included relationships
Seen on chameleon-api, when a table (let's say `comments`) has a column named _type_, when making a request that includes said table (`users?include=comments`).
The included comments don't have the field type.

The code responsible is this function:
https://github.com/ebryn/jsonapi-ts/blob/4a907b9086ae766245cdf5003f0f13149838b21e/src/application.ts#L133

spersico:
This PR, when merged, will close #87. Necessary for Chameleon.
Basically, this PR removes _"type"_ as a sort of *reserved word* for column names, by using the schema definition instead of the replaced type property.
It also, is a little more efficient, as we now go through the array just once.

Oh, and it also removes an *Any* type definition.


src/resource.ts
```diff
@@ -14,7 +14,7 @@ export default class Resource {
   id?: string;
   type: string;
   attributes: ResourceAttributes;
-  relationships: ResourceRelationships;
+  relationships: ResourceRelationships | ResourceRelationships[];
```
joelalejandro: @spersico Careful, `relationships` is never an array:

```js
relatioships: {
  user: {
    data: [ {id: 1, type: "user"}, {id: 2, type: "user"} ]
  },
  address: {
    data: [ {id: 3, type: "address"}, {id: 4, type: "address"}, {id: 5, type: "address"} ]
  }
}
```

PR #89, opened 2019-04-22
Fix response to requests without include filters

spersico:
This PR transforms the response to requests of models with belongsTo relationships, to comply more with the JSONAPI specs.

Currently, hitting the API with a GET to the `/comments` url will respond with this:
``` 
{
   "data": [
        {
           "id": 1,
           "type": "comment",
           "attributes": {
                 "body": "bodybody3"
           },
        },
   ...
   ]
}
```
and the only way to see the attribute author_id is by adding it to the attributes list, and that results in:

``` 
{
   "data": [
        {
           "id": 1,
           "type": "comment",
           "attributes": {
                 "body": "bodybody3",
                 "author_id": 1
           },
        },
   ...
   ]
}
```
which is not JSONAPI compliant.
So, by passing through the relationship keys at the knexProcessor level, and then serializing them at the application level, that same request gets this response:
``` 
{
   "data": [
        {
           "id": 1,
           "type": "comment",
           "attributes": {
                 "body": "bodybody3"
           },
            "relationships": {
                "author": {
                    "id": 1,
                    "type": "user"
                }
            }
        },
   ...
   ]
}
```

To ensure that chameleon doesn't break horribly because of this (I'm aware that chameleon has no relationships defined, and the relationship keys are loaded as attributes) , I added the possibility of not removing the properties from the attributes, if it's loaded in the attributes list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebryn/jsonapi-ts",
-  "version": "0.1.20",
+  "version": "0.1.21",
   "description": "",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/src/application.ts
+++ b/src/application.ts
@@ -99,17 +99,51 @@ export default class Application {
 
 
     return included.length ?
-      { included: uniqueIncluded, data: this.serializeResources(data) } :
-      { data: this.serializeResources(data) };
+      { included: uniqueIncluded, data: await this.serializeResources(data) } :
+      { data: await this.serializeResources(data) };
   }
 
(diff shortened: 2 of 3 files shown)


joelalejandro:
@ebryn Would you mind taking a look at this? I get the feeling this isn't exactly right. If I recall correctly, we decided that if no `include` parameter is set in the URL, no relationship data whatsoever (including IDs) should be rendered.

ebryn:
@joelalejandro I think you misunderstood. I believe I was referring to `included` relationship data not the `relationships` object itself.

PR #90, opened 2019-04-22
Fix reflexive relationships

### LINKED ISSUE #85: Reflexive relations
Reflexive relations 0 .. 1 between the same entity are not currently supported.

Example schema 

```ts
export default class Comment extends Resource {
  static get type() {
    return 'comment';
  }

  static schema = {
    attributes: {
      body: String,
      created_at: String,
      updated_at: String,
      parentCommentId: String
    },
    relationships: {
      user: {
        type: () => User,
        belongsTo: true,
        foreignKeyName: "userId"
      },
      parentComment: {
        type: () => Comment,
        belongsTo: true,
        foreignKeyName: "parentCommentId"
      },
      thread: {
        type: () => Thread,
        belongsTo: true,
        foreignKeyName: "threadId"
      }
    }
  };
}
```

It errors out when I do this query: `/comments?include=parentComment` with this error output:

`JSONAPI-TS:  { error: table name "comments" specified more than once `

spersico:
This PR should be merged after #88, and closes #85 . As I don't know the actual order in which this PR, #88, and #87 are going to be merged, I left it as it is, and will fix it when it needs to.

----------------------
**UPDATE APRIL 25:**
Added fix so that **ALL** relationships are shown, even when they are not set, or a include filter is present.

image

-----------------------
This PR only fixes reflexive relationships where the FK is on the base model (I didn't touch the hasMany SQL query generator).

------------

Basically, this PR does two things:
- It changes the _belongsTo_ SQL generator, so that one of the tables has an alias, to avoid a DB error.
- It adds a filter to the **serializer**, because it seems that the change in the SQL makes possible the existence of undefined objects in the **data.relationships** property.

This is an example of the endpoint working
image


src/application.ts
```diff
@@ -155,7 +155,7 @@ export default class Application {
     const resourceClass = await this.resourceFor(data.type);
     const schemaRelationships = resourceClass.schema.relationships;
 
-    Object.keys(data.relationships).forEach(relationshipName => {
+    Object.keys(data.relationships).filter(relationshipName => data.relationships[relationshipName]).forEach(relationshipName => {
```
joelalejandro: @spersico These filters could simply be `.filter(Boolean)`

src/processors/knex-processor.ts
```diff
@@ -255,12 +255,12 @@ export default class KnexProcessor<
       return query
         .join(
           this.tableName,
-          `${foreignTableName}.id`,
+          `base${foreignTableName}.id`,
```
joelalejandro: @spersico What does `base` mean? I know it works as a prefix, but I'd like to know why do we call it that way.

spersico: because.... I felt like it :-P

src/application.ts
spersico: okay, I'll give it a go

src/processors/knex-processor.ts
spersico: I'll give it a more descriptive name, like... **other** or something like that, I'll check the generated sql string and see

src/application.ts
spersico: @joelalejandro in https://github.com/ebryn/jsonapi-ts/pull/90/files#diff-8f42631f8d7d8cbaf47074d3ea5536d1R173 that worked, but here it breaks. Surely because one gets the keys (Object.keys), and the other the values (Object.values).
Even then, remember that this PR overlaps the #88, so when that one gets merged, part of this code will change.
I'll re-review it again then.

joelalejandro: It's not that important, we can iterate on this later, @spersico.

ebryn:
@spersico Can you fix these merge conflicts?

spersico:
> @spersico Can you fix these merge conflicts?

yep, I was expecting this. I'll see to it later

PR #93, opened 2019-04-26
adds resource deserializer to requests

### LINKED ISSUE #86: Relations are ignored when creating resources
[POST]
```js
{
  "data": {
    "type": "comments",
    "attributes": {
      "body": "New message"
    },
    "relationships": {
      "user": {
        "data": { "type": "user", "id": "1" }
      }
    }
  }
}
```

Ignores the user relation completely, currently the work around is to add the `userId: 1` to the attributes hash

[POST]
```js
{
  "data": {
    "type": "comments",
    "attributes": {
      "body": "New message",
      "userId": "1"
    },
    "relationships": {
      "user": {
        "data": { "type": "user", "id": "1" }
      }
    }
  }
}
```

spersico:
closes #86
This PR adds de-serialization to requests, to read **relationship** objects passed. This is done by transforming the objects passed in the request into the FK of the resource to be added/updated in the database.

Basically, previous to this PR, any **POST** or **PATCH** that had this payload in the request, ignored the specified relationships
```
{
    "data": {
        "type": "user",
        "attributes": {
            "email": "flyster91@prototypal.io"
        },
        "relationships": {                                             // ignored
        	"articles":{                                           // ignored
        		"data":{                                       // ignored
        			"id": 2,                               // ignored
        			"type":"article"                       // ignored
        		}                                              // ignored
        	}                                                      // ignored
        }                                                              // ignored
    }
}
```
-----------
## Warning
For now, this PR limits it's scope to belongsTo relationships, as hasMany relationships is a much more complex endeavor (as one add operation could involve multiple update or create, or even delete operations), and requires further analysis.

### DIFF
--- a/src/application.ts
+++ b/src/application.ts
@@ -4,9 +4,7 @@ import Resource from "./resource";
 import {
   Operation,
   OperationResponse,
-  Links,
-  ResourceRelationships,
-  ResourceRelationship
+  ResourceRelationshipData
 } from "./types";
 import pick from "./utils/pick";
 import unpick from "./utils/unpick";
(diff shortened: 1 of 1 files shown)


src/application.ts
```diff
@@ -51,14 +49,36 @@ export default class Application {
...
   }
 
   async createTransaction(ops: Promise<OperationResponse>[]) {
     return await Promise.all(ops);
   }
 
+  async deserializeResource(op: Operation) {
+    const resourceClass = await this.resourceFor(op.ref.type);
+    const schemaRelationships = resourceClass.schema.relationships;
+    if (op.data && op.data.attributes) {
```
joelalejandro: Something I usually do for easier readibility: If a condition is relevant to a large portion of a function's code, you can negate it and do an earlier exit:

```ts
if (!op.data || !op.data.attributes) {
  return op;
}

// Code that works with the passing condition.
// yadayadayada...

return op;
```

Yes, it has the "downside" of a _double return_. I'll leave it up to you to decide if this works here or not.

spersico: You're right, I'll fix it right away

PR #94, opened 2019-04-26
Serialized included resources

### LINKED ISSUE #92: Included Relationship Resources aren't valid JSONAPI
Expands upon #81 .

#### Basically, this is about enforcing the same resource structure to all resource shown in responses.

Included Resources aren't being return as valid JSONAPI resources, just extracted into the included property of the response.

For example, the response to a `GET /comments/3?include=parentComment` is:
image

when it should be:
image

The function responsible for this is:
https://github.com/ebryn/jsonapi-ts/blob/d352cef3e19294d8da2ce329b02db7773d7fd627/src/application.ts#L180

spersico:
closes #92 
This PR basically applies the same transformation suffered for the requested resources, in which the foreign keys are transformed into resource references.

Or with an example, it transform this part of a response:
image

into this
image


src/application.ts
```diff
@@ -111,10 +106,15 @@ export default class Application {
     }
 
     if (Array.isArray(data)) {
-      return Promise.all(data.map(record => this.serializeResources(record)));
+      const resource = await this.resourceFor(data[0].type);
```
joelalejandro: @spersico This might not work if you have different included types. Suppose you request a list books, including `authors` and `sellers`. If I'm reading this correctly, you're gonna be assuming every resource is an `author`.

```ts
// Minor enhancement to always work with arrays.
const arraify = (val) => Array.isArray(val) ? val : [val]:

return Promise.all(
  arraify(data).map(async record => 
    this.serializeResource(record, await this.resourceFor(record.type))
  )
);
```

spersico: > @spersico This might not work if you have different included types. Suppose you request a list books, including authors and sellers. If I'm reading this correctly, you're gonna be assuming every resource is an author.

This function (serializeResources(data)) affects only to the main resource of the operation, meaning, that for your example this would only affect books. 
What it IS true, tough, is that I assume that an operation has only one main resource type, so the data parameter would only have ONE type of resource (the included resources are handled separately).

spersico: I'm wrong in assuming that?

spersico: I pushed the arraify approach, I liked that :-)

joelalejandro: Nope, your assumption is correct :) Glad you liked the arraify thingy

PR #95, opened 2019-04-29
Custom PK support

### LINKED ISSUE #80: We assume the DB uses a column named `id`
We should be able to configure that per resource. For example, on one of our projects it's called `Id`

spersico:
closes #80 .

----------------
#### This PR should be merged AFTER my previous ones, as this touches a fair amount of files.

-------------

So far, this PR adds support for CRUD in a Resource which has a new schema attribute: _primaryKeyName_.

For example, a resource could be defined like this:
image

A GET to `/votes/1` brings this:
image

And a POST to `/votes` request and response would be:
image


src/application.ts
```diff
@@ -193,7 +193,7 @@ export default class Application {
     Object.keys(data.relationships).forEach(relationshipName => {
       const relationships = this.serializeRelationship((data.relationships[
         relationshipName
-      ] as unknown) as Resource | Resource[]);
+      ] as unknown) as Resource | Resource[], schemaRelationships[relationshipName].type().schema.primaryKeyName);
```
joelalejandro: @spersico Can we create some variables to make these lines a bit more... readable? 🙏

src/processors/knex-processor.ts
```diff
@@ -147,18 +151,18 @@ export default class KnexProcessor<
...
+      .where({ [primaryKeyName]: id })
       .select(getColumns(this.resourceClass))
       .first();
   }
 
   async add(op: Operation): Promise<HasId> {
     const { type } = op.ref;
-
-    const ids = await this.getQuery().insert(op.data.attributes, "id");
+    const primaryKeyName = this.resourceClass.schema.primaryKeyName || "id";
```
joelalejandro: @spersico Seems like we could make the `schema` assume `id` has default so this defaulting can be done only once?

src/processors/knex-processor.ts
```diff
@@ -245,24 +249,32 @@ export default class KnexProcessor<
...
     const query = relationProcessor.getQuery();
     const foreignTableName = relationProcessor.tableName;
     const sqlOperator = Array.isArray(result) ? "in" : "=";
+
+    const primaryKey = this.resourceClass.schema.primaryKeyName || "id";
+
     const queryIn: string | string[] = Array.isArray(result)
-      ? result.map((a: Resource) => a.id)
-      : result.id;
+      ? result.map((a: Resource) => a[primaryKey])
```
joelalejandro: @spersico can we name this variable `resource`? `a` means nothing in this context...

src/application.ts
```diff
@@ -53,6 +54,7 @@ export default class Application {
     processor: OperationProcessor<Resource>
   ): Promise<OperationResponse> {
     const result = await processor.execute(await this.deserializeResource(op));
+    console.log('result', result)
```
joelalejandro: @spersico This should probably go away

src/application.ts
```diff
@@ -74,10 +76,8 @@ export default class Application {
           op.data.relationships.hasOwnProperty(relName)
       )
       .reduce((relationAttributes, relName) => {
-        let key =
-          schemaRelationships[relName].foreignKeyName ||
-          `${schemaRelationships[relName].type().type}Id`;
-        let value = (<ResourceRelationshipData>(
+        const key = schemaRelationships[relName].foreignKeyName || `${relName}Id`;
```
joelalejandro: @spersico In another PR we should set a function to build `${relName}Id`, so can have in a single place the builder for that string.

src/application.ts
```diff
@@ -189,13 +182,18 @@ export default class Application {
...
-
-    Object.keys(data.relationships).forEach(relationshipName => {
-      const relationships = this.serializeRelationship((data.relationships[
-        relationshipName
-      ] as unknown) as Resource | Resource[], schemaRelationships[relationshipName].type().schema.primaryKeyName);
-      data.relationships[relationshipName] = {
-        data: relationships
+    Object.keys(data.relationships).forEach(relName => {
+      const fkName = schemaRelationships[relName].belongsTo
+        ? "id"
```
joelalejandro: @spersico Shouldn't this use `DEFAULT_PRIMARY_KEY`?

src/application.ts
```diff
@@ -225,7 +223,7 @@ export default class Application {
     if (!data) {
       return null;
     }
-
+    console.log('data', data)
```
joelalejandro: @spersico This should probably go away

src/application.ts
spersico: Noup, in line 168 we build the resources object using "id" from the proper FK. 
This forEach ensures only that all elements in the relationships object have ONLY an **id** and **type** attributes, inside a data object.

src/application.ts
spersico: created an issue with this

PR #96, opened 2019-04-30
Implements per-request application instances

### LINKED ISSUE #13: Add ApplicationInstance to instantiate requests
For now `Application` is a singleton which can cause an issue eg. with static current user

joelalejandro:
Resolves #13.

This PR changes the way the processors interact with the `Application` object, by introducing an intermediate `ApplicationInstance` object, which holds the `user` for a particular request, and a reference to the `Application` definition where everything takes place.

This will allow us to scope more code in a per-request level, without further need to keep dumping stuff in the `Application` object.

> ℹ️ Some changes are due to prettifying old code.


src/application.ts
```diff
@@ -96,14 +99,14 @@ export default class Application {
     const processor = processors.find(p => p !== false);
 
     if (processor) {
-      return new processor(this);
+      return new processor(new ApplicationInstance(this));
```
rtablada: I don't think this is actually doing things correctly?

Seems like you're getting a new application instance per processor not per bulk...?

joelalejandro: Yeah, this might not be a good idea.

src/application.ts
```diff
@@ -33,10 +33,12 @@ export default class Application {
   }
 
   async executeOperations(ops: Operation[]): Promise<OperationResponse[]> {
+    let applicationInstance = new ApplicationInstance(this);
+
     return await this.createTransaction(
       ops
         .map(async op => {
-          const processor = await this.processorFor(op.ref.type);
+          const processor = await this.processorFor(op.ref.type, applicationInstance);
```
ebryn: should processorFor move to the instance so we don't have to pass it?

joelalejandro: Yeah, we can change that.

PR #109, opened 2019-05-03
Authentication/Authorization fixes

spersico:
This PR fixes authorization. And it adds a whole bunch of models, utils and processors to the dummy app so we can test this in the future.
It also handles better the authentication process, and the failed search of a model.


ebryn:
@joelalejandro please review

PR #125, opened 2019-05-06
Fixes several serializer issues

spersico:
Remaining:
- [x] Ensure that passwords aren't leaked as included resources
- [x] Export string utils
- [x] Test (I didn't) ADD and PATCH operations with hasMany and BelongsTo relationships (to test the deserializer)
- [x] Add fallback value to the serializer
- [x] And
image


tests/dummy/src/resources/user.ts
```diff
@@ -12,17 +12,17 @@ export default class User extends JsonApiUser {
     },
 
     relationships: {
-      articles: {
+      article: {
```
ebryn: this should be pluralized since it's a has many

tests/dummy/src/resources/user.ts
```diff
@@ -12,17 +12,17 @@ export default class User extends JsonApiUser {
...
         type: () => Article,
         hasMany: true,
-        foreignKeyName: "authorId"
+        foreignKeyName: "author_id"
       },
-      comments: {
+      comment: {
```
ebryn: this should be pluralized since it's a has many

joelalejandro:
@ebryn Fixed. It was just a declaration mistake, no changes to the serialization process were required.

PR #134, opened 2019-05-08
WIP serialize json string for inserts

rtablada:


rtablada:
@spersico 🤔 that is a problem. It should come back json parsed if it goes in json stringified (no matter what the value).

spersico:
@rtablada I believe that for that, you need to look into 
https://github.com/ebryn/jsonapi-ts/blob/e66542a176cebd0e5ede1b73d0e8310189ed8a26/src/processors/operation-processor.ts#L54
so that, while serializing the response, you JSON.parse the value, according to the schema

rtablada:
I’m gonna make an acceptance test and try it against postgres we should have null in null out for real values. I don’t care so much what the value is in the DB as long as it doesn’t error.

rtablada:
I'm working on adding the acceptance test. Need to check the behavior of PG vs MSSQL vs MySQL vs SQLite

PR #136, opened 2019-05-08
Fix sort clause

spersico:
**_ugh, I started from not-yet merged branch, so this a 2x1 PR_**
This PR fixes a line that got overwriten. That allows for the serialization of custom FK belongsTo relationships.

**_And for the same price, ladies and gentlemen!_**

It also removes the semi-harcoded "serialization" of fields in the sort filter, and puts it through the proper serializator. And it adds an exception for customFKfields.


spersico:
If part of this PR has a problem, let me know, I can divide this PR in two.

PR #139, opened 2019-05-08
Fixed bad token serialization and prevents password leaking in JWT

joelalejandro:
The built-in SessionProcessor was not serializing the `User` model before JWT'ing it, and it was also not removing sensitive data fields such as Password-type data.

### DIFF
--- a/src/processors/session-processor.ts
+++ b/src/processors/session-processor.ts
@@ -35,18 +35,25 @@ export default class SessionProcessor<T extends Session> extends KnexProcessor<T
       throw jsonApiErrors.AccessDenied();
     }
 
-    const userId = String(user.id);
+    const userPk = userType.schema.primaryKeyName || DEFAULT_PRIMARY_KEY;
+    const passwordFields = Object.keys(userType.schema.attributes).filter(
+      attribute => userType.schema.attributes[attribute] === Password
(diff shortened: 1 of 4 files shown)


src/processors/session-processor.ts
```diff
@@ -35,18 +35,25 @@ export default class SessionProcessor<T extends Session> extends KnexProcessor<T
...
 
-    const userId = String(user.id);
+    const userPk = userType.schema.primaryKeyName || DEFAULT_PRIMARY_KEY;
+    const passwordFields = Object.keys(userType.schema.attributes).filter(
+      attribute => userType.schema.attributes[attribute] === Password
+    );
+    const userId = String(user[userPk]);
 
-    delete user.id;
+    [userPk, ...passwordFields].forEach(field => delete user[field]);
```
spersico: Wouldn't it be better to destructure the user into the props you want?
Something like 
`const { id, ...passwordFields, ...cleanUser} = user;`
and then just use the cleanUser

joelalejandro: That looks better, actually. I'll refactor it that way.

joelalejandro: Resolved in #140.

PR #140, opened 2019-05-09
Adds support for Addons, refactors built-in auth into UserManagementAddon

### LINKED ISSUE #138: Figure out a better API for importing User/Session
For virtually any serious implementation of the User/Session resources, the built-in objects in JSONAPI-TS are too simple and hard to override without falling in excessive repetition.

There should be some sort of middleware for the `App` object that allows to configure these resources, kinda like a plugin system of sorts.

joelalejandro:
Merge after #139.

This PR implements a basic architecture for extending the app's types and processors via _addons_.

An `Addon` is basically an extension of a class with the following API:

```ts
class Addon implements IAddon {
  constructor(public readonly app: Application, public readonly options?: AddonOptions) {
    this.app = app;
    this.options = options;
  }

  async install() {
    /* Addon, do your magic here 👲 */
  }
}
```

The problem with the previous approach for implementing User/Session, as described in #138, was that if you made any change to the User resource, you'd need to reimplement both the User and Session processors to re-bind the resource.

The `UserManagementAddon` takes care of that for you, with a simple invocation:

```ts
app.use(UserManagementAddon, {
  userResource: User
});
```

I also moved the previously-known `login` and `password` services to this addon, in the form of callbacks in the configuration object:

```ts
export type UserManagementAddonOptions = AddonOptions & {
  userResource: typeof User;
  userEncryptPasswordCallback?: (op: Operation) => Promise<ResourceAttributes>;
  userLoginCallback?: (op: Operation, userDataSource: ResourceAttributes) => Promise<boolean>;
  userGenerateIdCallback?: () => Promise<string>;
  usernameRequestParameter?: string;
  passwordRequestParameter?: string;
};
```

You'll notice, this way, the dummy app has no further need of a session resource/processor or an user processor -- it only requires a `User` resource because, as expected, it introduces its own data (beyond what the built-in, placeholder object offers).

This architecture is decoupled enough to be used for any non-core features we want to develop in the future as well.

This resolves #138.


src/addons/user-management.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.app.types.push(this.options.userResource, sessionResourceType);
+    this.app.processors.push(this.createUserProcessor(), this.createSessionProcessor(sessionResourceType));
+  }
+
+  private createUserProcessor() {
+    return (options =>
+      class UserProcessor<T extends User> extends JsonApiUserProcessor<T> {
+        public static resourceClass = options.userResource;
+
+        protected async generateId() {
```
spersico: this should, somehow, be optional.
Unless I'm looking at it wrong, we should support automatic Id generation (which is fairly common) from the database.

joelalejandro: @spersico It is optional! :D

See the signature of the options object:

```ts
export type UserManagementAddonOptions = AddonOptions & {
  userResource: typeof User;
  userEncryptPasswordCallback?: (op: Operation) => Promise<ResourceAttributes>;
  userLoginCallback?: (op: Operation, userDataSource: ResourceAttributes) => Promise<boolean>;
  // This question mark next to the name means "optional".
  userGenerateIdCallback?: () => Promise<string>;
  usernameRequestParameter?: string;
  passwordRequestParameter?: string;
};
```

spersico: uh, sorry
In my defense, I was sleepy, and had no coffee

src/application.ts
```diff
@@ -40,10 +44,20 @@ export default class Application {
...
+    this.addons = [];
     this.serializer = new (settings.serializer || JsonApiSerializer)();
   }
 
+  use(addon: typeof Addon, options?: AddonOptions) {
+    if (this.addons.find(installedAddon => installedAddon.addon === addon)) {
+      return;
+    }
+
+    new addon(this, options).install().then(() => {
```
ebryn: why is `then` used?

src/index.ts
```diff
@@ -36,7 +38,11 @@ export {
   SessionProcessor,
   User,
   Session,
-  Password
+  Password,
+  // Addons
```
ebryn: comment?

joelalejandro: It's a valid comment, not a commented import.

src/application.ts
joelalejandro: @ebryn Because `app.use` is typically going to be used on a root scope and it would be unconfortable to use it as an async function, from a developer ergonomics standpoint.

PR #149, opened 2019-05-16
Converts error objects to Error instances

### LINKED ISSUE #110: JsonApiError objects clutter the logs with non-error promise rejection warnings
Since we're using POJOs to represent errors in JSONAPI-TS, Node is throwing the following warning every time a JsonApiError is thrown in an async function:

```
(node:1626) Warning: a promise was rejected with a non-error: [object Object]
```

We should consider refactoring the error system to work with actual Error objects that can be converted into JsonApiError-like POJOs.

joelalejandro:
Fixes #110.


joelalejandro:
That level of granularity is for another PR :) we'll get there.

Issue #107, opened 2019-05-03
Update documentation


joelalejandro:
Outdated sections:

- [x] Relationships (#153)
- [x] Schema definition (need to talk about primaryKeyName) (#153)
- [x] Authentication/Authorization (#153)

Missing sections:

- [x] Bulk operations (#151)
- [x] WebSocket support  
  - [x] Using the jsonApiWebSocket handler (#151)
  - [x] Authenticating on WebSockets (#159)
- [x] Addons (#152)
- [x] Serialization (#160)

PR #153, opened 2019-05-16
Documentation: describes the UserManagement addon

joelalejandro:
Contributes to #107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,7 +917,7 @@ In order to put all the pieces together, JSONAPI-TS provides an [addon](#what-is
 
 You'll need to define at least two functions:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1015,6 +911,97 @@ The `IfUser()` helper syntax is as follows:
 IfUser(attributeName: string, attributeValue: string | number | boolean);
 ```
 
+### Using the `UserManagement` addon
+
+In order to put all the pieces together, JSONAPI-TS provides an [addon](#what-is-an-addon) to manage both user and session concerns.
+
+You'll need to define at least two functions:
```
spersico: I would make it clearer when both of these functions are used.
Something like:
"An `encryptPassword` callback that get's used in user creation process to set the password/security field"
and 
"A `login` callback that get's used in the login process"

I don't think that it's clear enough as it is to a just-arrived dev to understand

joelalejandro: Will improve per your suggestions :)

PR #158, opened 2019-05-20
add custom primary key, and relationships

spersico:
Contributes to #107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ A relationship should be defined on its two ends. For example, on the example, w
 
 ```
   static schema = {
-   attributes:{...},
+   attributes: { /* ... */ },
     relationships: {
       // ...,
       books: {
--- a/README.md
+++ b/README.md
@@ -90,11 +90,7 @@ This is a TypeScript framework to create APIs following the [1.1 Spec of JSONAPI
 3. Create an Application and inject it into your server. For example, let's say you've installed Koa in your Node application and want to expose JSONAPI via HTTP:
 
    ```ts
-   import {
-     Application,
-     jsonApiKoa as jsonApi,
-     KnexProcessor
-   } from "@ebryn/jsonapi-ts";
+   import { Application, jsonApiKoa as jsonApi, KnexProcessor } from "@ebryn/jsonapi-ts";
    import Koa from "koa";
 
    import Author from "./resources/author";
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,18 +189,59 @@ export default class Book extends Resource {
...
   };
 }
 ```
 
+#### Relationship Declarations
+
+Any number of relationships can be declared for a resource. Each relationship must have a type function which returns a Class, the kind of relationship, which can be **belongsTo** or **hasMany**, and
+The expected foreign key depends on the serializer, which can be customized, but with the default serializer, the FK is `${relationshipName}${primaryKeyName}`.
+
+A relationship should be defined on it's two ends. For example, on the example, with the above code in the _Book_ resource definition, a GET request to `books?include=author`, would include in the response the related user for each book, but for the inverse filter, in the _User_ resource schema definition, we should include:
```
joelalejandro:
```suggestion
A relationship should be defined on its two ends. For example, on the example, with the above code in the _Book_ resource definition, a GET request to `books?include=author`, would include in the response the related user for each book, but for the inverse filter, in the _User_ resource schema definition, we should include:
```

README.md
```diff
@@ -191,18 +189,59 @@ export default class Book extends Resource {
...
+
+```
+  static schema = {
+   attributes:{...},
```
joelalejandro:
```suggestion
   attributes: { /* ... */ },
```

README.md
```diff
@@ -191,18 +189,59 @@ export default class Book extends Resource {
...
+    relationships: {
+      ...,
```
joelalejandro:
```suggestion
      // ...,
```

README.md
```diff
@@ -191,18 +189,59 @@ export default class Book extends Resource {
...
+        hasMany: true,
+        foreignKeyName: "authorId"
+      },
+    }
+  };
+```
+
+For a GET request to `users?include=books` to include the books related to each user.
+
+> Declaring a relationship is necessary to parse each resource and return a JSONAPI compliant response. Also, it gives the API the necessary information so that `include` filters work correctly.
```
joelalejandro:
```suggestion
> Declaring a relationship is necessary to parse each resource and return a JSONAPI-compliant response. Also, it gives the API the necessary information so that the `include` clause works correctly.
```

PR #167, opened 2019-05-23
Adds built-in roles and permissions

### LINKED ISSUE #166: Built-in roles and permissions
@rtablada suggested a simpler way, exposing two callback methods that the auth service can invoke to fill in the data about roles and permissions.

Original proposal:

> In order to provide more granular control in terms of endpoint authorization, a logical next step would be to expand the current `UserManagementAddon` with two new resources:
> 
> - **Role:** represents a group of permissions. A User can have many roles.
> 
> - **Permissions:** represents a capability to perform a certain action. Can belong to multiple roles. A User can have multiple permissions.
> 
> We'd also add new helpers to match these resources to a User: `IfUserHasRole`, `IfUserHasEveryRole`, `IfUserDoesNotHaveRole`, `IfUserHasPermission`, `IfUserHasEveryPermission`, `IfUserDoesNotHavePermission`.
> 
> What should the developer do in order to take advantage of this new feature?
> 
> 1) Use the built-in new resources or extend and pass them to the options for the `UserManagementAddon`.
> 
> 2) Use the *Role and *Permission helpers along with `@Authorize`.
> 
> Both `Role` and `Permission` resources have a `name: string` attribute and a `granted: boolean` attribute. Like any other resource, they can be handled by custom processors to adapt to pre-existing databases or any custom logic the developer wants to implement.

joelalejandro:
Resolves #166.

Usage:

1. Define providers for roles and permissions.

```ts
app.use(UserManagementAddon, {
  // ...
  userRolesProvider: async (user) => {
    // Do some magic to get the role names.
    return [...];
  },
  userPermissionsProvider: async (user) => {
    // Do some magic to get the permission names.
    return [...];
  }
  // ...
} as UserManagementAddonOptions);
```

2. Check for the presence of a given role/permission with the  `@Authorize` decorator:

```ts
@Authorize(IfUserHasRole("Admin"))
public async add(op: Operation) {
   // Admin-only operation. Other roles will receive a 403.
}
```

### DIFF
--- a/src/decorators/if-user.ts
+++ b/src/decorators/if-user.ts
@@ -4,7 +4,7 @@ export function ifUser(attribute: string, value: string | number | boolean | str
   return { attribute, value };
 }
 
-export function ifUserNotMatches(
+export function ifUserDoesNotMatches(
   attribute: string,
   value: string | number | boolean | string[] | number[]
 ): AttributeValueMatch {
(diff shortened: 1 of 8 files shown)


src/decorators/if-user.ts
```diff
@@ -4,7 +4,7 @@ export function ifUser(attribute: string, value: string | number | boolean | str
   return { attribute, value };
 }
 
-export function ifUserNotMatches(
+export function ifUserDoesNotMatches(
```
spersico: THANKS!

PR #172, opened 2019-05-24
Documentation: Roles and permissions

### LINKED ISSUE #168: Add documentation for roles and permissions
Once the API has been validated, document it.

joelalejandro:
Resolves #168.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,9 @@ This is a TypeScript framework to create APIs following the [1.1 Spec of JSONAPI
 - [Authentication and authorization](#authorization)
   - [Defining an `User` resource](#defining-an-user-resource)
   - [Using the `@Authorize` decorator](#using-the-authorize-decorator)
-  - [Using the `IfUser()` helper](#using-the-ifuser-helper)
   - [Using the `UserManagement` addon](#using-the-usermanagement-addon)
   - [Configuring roles and permissions](#configuring-roles-and-permissions)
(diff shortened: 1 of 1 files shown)


joelalejandro:
🙄 I forgot to explain the new Roles/Permission helpers... 🤦‍♂

README.md
```diff
@@ -1169,6 +1170,45 @@ app.use(UserManagementAddon, {
...
+
+`role-provider.ts`
+```ts
+import { ApplicationInstance, User } from "@ebryn/jsonapi-ts";
+
+export default async function roleProvider(this: ApplicationInstance, user: User): Promise<string[]> {
+  const userRoleProcessor = this.processorFor("userRole");
+  
+  return (
+    await roleProcessor.getQuery()
```
spersico: Everything else it's great, but getQuery is not defined nor documented anywhere (I checked, we have another example that uses it: https://github.com/ebryn/jsonapi-ts/blob/master/README.md#using-computed-properties-in-a-processor ).
I believe that we should explain what does getQuery does, what can be used for, and what's it's output, so that when we use it in examples, the reader has a reference to what it does.

spersico:
> 
> 
> 🙄 I forgot to explain the new Roles/Permission helpers... 🤦‍♂

oh! yeah :-P

README.md
joelalejandro: I'll add a separate issue for that. It's part of the `KnexProcessor` API.

spersico: great!

Issue #188, opened 2019-05-28
Pagination query is not working


spersico:
Following the docs: https://github.com/ebryn/jsonapi-ts#get-operations

I tried to paginate the results of the articles collection (in the dummy app). Couldn't make it work:
![image](https://user-images.githubusercontent.com/10502605/58452544-1be3da80-80ee-11e9-950c-20715b1edebd.png)

Issue #187, opened 2019-05-28
Fields filter not working


spersico:
Following the docs: https://github.com/ebryn/jsonapi-ts#get-operations

I tried the *fields* filter, doesn't work. Example below:
![image](https://user-images.githubusercontent.com/10502605/58451625-97438d00-80ea-11e9-8470-a252fe9ecd6d.png)

Issue #186, opened 2019-05-28
Restricted resources can be retrieved even with unauthenticated requests


spersico:
I'm able to surpass a restricted operation, by including the restricted resource if the restricted resource is referenced by another resource.
For example, in the dummy app:

With a custom User Processor, which restricts the get operation:
`@Authorize(IfUserDoesNotHaveRole("foo"))
  public async get(op: Operation) {
    return super.get(op);
  }`

I'm able to get a user like this:
![image](https://user-images.githubusercontent.com/10502605/58445131-c0a3ef00-80d1-11e9-833d-39551d21a2a6.png)

joelalejandro:
@spersico It's correct for you to get the resource, since the user does not have the "foo" role. However, the bug is confirmed, since changing the settings for `@Authorize` reproduced it.

joelalejandro:
The problem is the way we include resources right now does not make use of the `get` method for their bound processors, so the `@Authorize` decorator never gets to do its job.

PR #189, opened 2019-05-28
More tests!

### LINKED ISSUE #6: Write CRUD acceptance tests
@rtablada has added some basic infrastructure for tests.

Let's write tests that exercise CRUD operations via HTTP & the bulk operation endpoint

Essentially we'll make HTTP requests and assert their responses look as expected

spersico:
part of #6 
It includes a bunch more filter tests (the like and like with multiple comparers in a field included)

Some bugs were found and reported
 ● Comments › GET › Get Comment by id - Only show the body attribute ( #187 )
 ● Comments › GET › Get Comments - Test pagination and sorting - Should only show one (#188)
 ● Articles › GET › UNAuthenticated - Get an specific article with it's author - Should fail (#186)

### DIFF
--- a/tests/data/migrations/20190506190058_add_articles_table.ts
+++ b/tests/data/migrations/20190506190058_add_articles_table.ts
@@ -5,7 +5,7 @@ export async function up(knex: Knex): Promise<any> {
     table.increments("id").primary();
     table.string("body");
     table
-      .string("author")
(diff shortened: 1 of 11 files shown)


spersico:
@joelalejandro currently 3 tests are failing, which are the ones corresponding to each created issue

PR #195, opened 2019-05-29
Allows to filter before updating/removing

### LINKED ISSUE #193: Make "update" and "remove" operations filter-aware
While we always use the `op.ref.id` to select what resource we're updating or removing, we can also benefit from adding filters to make sure the resource matches any conditions the client specifies.

This can be useful for multi-tenant applications, where a resource is scoped to the access level of a user.

joelalejandro:
Resolves #193.


spersico:
LOL

PR #201, opened 2019-06-03
Prevents access to unauthorized resources via `included`

### LINKED ISSUE #186: Restricted resources can be retrieved even with unauthenticated requests

joelalejandro:
This PR extracts the condition evaluation code of the `@Authorize` decorator into a separate function `conditionsPass` that is now shared with the original decorator and a new filtering function called `canAccessResource`.

This new function is now used by the `Application#buildOperationResponse` function to exclude from the `included` resources array any resources that aren't authorized by the conditions applied on their processors using `@Authorize`.

The trick is to decorate the function with a `Symbol` called `accessRules`, which copies the conditions passed to the decorator for further evaluation. Then, `canAccessResource` is called after `serializeIncludedResources` to remove any unauthorized resources when it returns `false`.

Resolves #186.

Also, it enables us to re-activate the test commented by #200.

### DIFF
--- a/src/decorators/authorize.ts
+++ b/src/decorators/authorize.ts
@@ -3,61 +3,72 @@ import { AttributeValueMatch, AttributeValue } from "../types";
 
 import decorateWith from "./decorator";
 import OperationProcessor from "../processors/operation-processor";
-import { Resource } from "..";
+import Resource from "../resource";
+import ApplicationInstance from "../application-instance";
 
 type PrimitiveValue = string | number | boolean | object;
 
 const match = (actual: AttributeValue) => (item: PrimitiveValue) => (actual as any[]).includes(item);
+const ACCESS_RULES = Symbol("accessRules");
+
+function conditionsPass(
+  appInstance: ApplicationInstance,
+  { attribute, value, operator = "some" }: AttributeValueMatch
+) {
+  const actual: AttributeValue = appInstance.user.attributes[attribute];
+  const expected = value;
+
+  if (Array.isArray(actual)) {
+    if (Array.isArray(expected)) {
+      if (operator === "some") {
+        return expected.some(match(actual));
+      }
+
+      if (operator === "every") {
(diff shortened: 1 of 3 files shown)


src/decorators/authorize.ts
```diff
@@ -3,61 +3,72 @@ import { AttributeValueMatch, AttributeValue } from "../types";
...
-  return function(this: OperationProcessor<Resource>) {
-    if (!this.appInstance.user) {
-      throw JsonApiErrors.Unauthorized();
-    }
+function conditionsPass(
+  appInstance: ApplicationInstance,
+  { attribute, value, operator = "some" }: AttributeValueMatch
+) {
+  const actual: AttributeValue = appInstance.user.attributes[attribute];
+  const expected = value;
```
spersico: this could be done directly by destructuring :-)

spersico: (I'm talking about the expected constant)

joelalejandro: Yes, but the typed syntax looks awful:

```ts
const [actual]: [actual: AttributeValue] = appInstance.user.attributes;
```

spersico: ```{ attribute:expected, value, operator = "some" }: AttributeValueMatch```

joelalejandro:
I'm re-enabling the test, will push a new commit in a bit.

src/decorators/authorize.ts
joelalejandro: This is actually wrong, we shouldn't destructure here. `attribute` is a dynamic name.

PR #202, opened 2019-06-03
Fix 404 user token

### LINKED ISSUE #194: We're returning 404 when a token is valid but contains a non-existing user
It should return an `invalid_token` error.

spersico:
The error was mainly in jsonapikoa. 
The process for an error to be shown in the response depends on jsonapikoa parsing that error and adding it to the response body.
The authentication had no way to show this error in the response.

The PR also fixes a related test badly written.
This PR, when merged, closes #194

### DIFF
--- a/src/middlewares/json-api-koa.ts
+++ b/src/middlewares/json-api-koa.ts
@@ -24,16 +24,23 @@ export default function jsonApiKoa(app: Application, ...middlewares: Middleware[
   const jsonApiKoa = async (ctx: Context, next: () => Promise<any>) => {
     const appInstance = new ApplicationInstance(app);
 
+    try {
+      await authenticate(appInstance, ctx);
+    } catch (error) {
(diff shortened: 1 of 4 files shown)


src/middlewares/json-api-koa.ts
```diff
@@ -82,18 +80,14 @@ async function handleJsonApiEndpoint(appInstance: ApplicationInstance, ctx: Cont
   if (["update", "remove"].includes(op.op) && !op.ref.id) return;
 
   try {
+    await authenticate(appInstance, ctx);
```
joelalejandro: @spersico Beware, I think this will make `/bulk` calls 100% unauthenticated. I think a better solution is to put this back where it was and wrap it there in a try/catch block.

spersico: ugh, I didn't saw that line, you are right.

PR #205, opened 2019-06-04
add more knex processor documentation

### LINKED ISSUE #176: Missing KnexProcessor API documentation
> [...] getQuery is not defined nor documented anywhere (I checked, we have another example that uses it: https://github.com/ebryn/jsonapi-ts/blob/master/README.md#using-computed-properties-in-a-processor ).
> I believe that we should explain what does getQuery does, what can be used for, and what's it's output, so that when we use it in examples, the reader has a reference to what it does.

_Originally posted by @spersico in https://github.com/ebryn/jsonapi-ts/pull/172_

spersico:
closes #176 
Upon discussion with @joelalejandro, we decided to not to document all the knexProcessor functions, as some don't make sense to use/extend, while extending the class behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -920,7 +920,7 @@ Apart from the operations, this processor has some other methods, that operation
 | `getQuery` | Returns an instance of [Knex.QueryBuilder](https://knexjs.org/#Builder), scoped to the table specified by `tableName` (the processor resource's data source) |
 | `tableName`     | Returns the table name for the resource handled by the processor                     |
 
-Using these two methods and the normal [knex functions](https://knexjs.org), you can extend however you want a processor.
+Using these two methods and the standard [Knex functions](https://knexjs.org/#Builder-wheres), you can extend a processor anyway you want to.
 
 ### Extending the `KnexProcessor` class
 
--- a/README.md
+++ b/README.md
@@ -918,7 +918,7 @@ Apart from the operations, this processor has some other methods, that operation
 | Method    | Description                                        |
 | --------- | ---------------------------------------------------- |
 | `getQuery` | Returns an instance of [Knex.QueryBuilder](https://knexjs.org/#Builder), scoped to the table specified by `tableName` (the processor resource's data source) |
-| `tableName`     | Returns the table name of the resource of the processor                     |
+| `tableName`     | Returns the table name for the resource handled by the processor                     |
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -913,6 +913,15 @@ It maps operations to queries like this:
 
 It receives a single argument, `options`, which is passed to the `Knex` constructor. See the [Knex documentation](https://knexjs.org/#Installation-client) for detailed examples.
 
+Apart from the operations, this processor has some other methods, that operations use (so tread carefully here), that you can use while making custom operations.
```
joelalejandro:
```suggestion
In addition to the operation handlers, this processor has some other methods that you can use while making custom operations. Note that all operations use these functions, so tread carefully here if you're interested in overriding them.
```

README.md
```diff
@@ -913,6 +913,15 @@ It maps operations to queries like this:
...
+
+| Method    | functionality                                        |
```
joelalejandro:
```suggestion
| Method    | Description                                        |
```

README.md
```diff
@@ -913,6 +913,15 @@ It maps operations to queries like this:
...
+| --------- | ---------------------------------------------------- |
+| `getQuery`| The common method to start a SQL request, starting with the table of the resource of the processor |
```
joelalejandro:
```suggestion
| `getQuery` | Returns an instance of [Knex.QueryBuilder](https://knexjs.org/#Builder), scoped to the table specified by `tableName` (the processor resource's data source) |
```

README.md
```diff
@@ -913,6 +913,15 @@ It maps operations to queries like this:
...
+| `tableName`     | Returns the table name of the resource of the processor                     |
```
joelalejandro:
```suggestion
| `tableName`     | Returns the table name for the resource handled by the processor                     |
```

README.md
```diff
@@ -913,6 +913,15 @@ It maps operations to queries like this:
...
+
+Using these two methods and the normal [knex functions](https://knexjs.org), you can extend however you want a processor.
```
joelalejandro:
```suggestion
Using these two methods and the standard [Knex functions](https://knexjs.org/#Builder-wheres), you can extend a processor anyway you want to.
```

PR #206, opened 2019-06-06
Express Support

### LINKED ISSUE #28: express support
We can port our Koa middleware to use Express pretty easily

spersico:
We now support Express!
The PR also refactors both the Express and Koa middlewares, extracting the shared functions into a separate http-utils file.

The tests now run both with Koa and Express. I added a simple function that makes switching a set of tests really easy.

I modified the docs to make sense of this new addition.

closes #28


spersico:
> 
> 
> Looks great! One tiny thing, though: do we need `yarn run:tests`? It's easy to confuse it with `yarn test`, I think.

um I left it to test the test app (if, for some reason, a test fails and we don't know why). It copies the run:dummy which runs the dummy app....
I could change the script name to run:test-app to avoid misunderstandings..

Issue #183, opened 2019-05-27
Cannot apply filters over computed properties


joelalejandro:
Using the dummy app, if you try to apply a filter over `voteCount` on the `article` resource, it'll fail because it tries to run the filter over the database.

spersico:
and... what should be the expected behaviour?

joelalejandro:
We can implement this in two stages, depending on other priorities and needs:

1) At the very least, don't mislead the user with a confusing error.

2) A great solution would be to have a native implementation of each comparer to enable the same behaviour databases offer on computed properties.

spersico:
Since #212, a more expressive and clear error is returned now.

PR #212, opened 2019-06-07
Error expressiveness improvement

spersico:
This PR solves the first goal expressed in #183, by giving a more clear error:
image

And it extends error creation, so that we (and users) can extend error creation to provide more details in each case.

### DIFF
--- a/src/errors/error.ts
+++ b/src/errors/error.ts
@@ -19,8 +19,6 @@ export default class JsonApiError extends Error implements IJsonApiError {
 
     this.status = status;
     this.code = code;
-    if (detail) {
-      this.detail = detail;
-    }
+    this.detail = detail;
   }
 }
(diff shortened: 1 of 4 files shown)


src/errors/error.ts
```diff
@@ -14,10 +14,13 @@ export default class JsonApiError extends Error implements IJsonApiError {
     about?: string;
   };
 
-  constructor(status: HttpStatusCode, code: string) {
+  constructor(status: HttpStatusCode, code: string, detail?: string) {
     super(`${status}: ${code}`);
 
     this.status = status;
     this.code = code;
+    if (detail) {
```
joelalejandro: @spersico Is there any need for the `if`?

spersico: noup, I thought that I needed it to avoid a "detail":undefined in the response, but it turns out that it's not necessary. I'll remove it

PR #221, opened 2019-10-20
Update Dependencies 

spersico:
I was tired of seeing the security warning.

I updated everything that I could without deleting the yarn.lock file, or updating jest to 24.9.0 (because knex migrations start failing like crazy and I don't want to deal with that).

It would be nice that someone else tries this before merging with master.
I've runned the test, built the package, and served the dummy app (and even tested a couple of endpoints manually) and it did not gave me any issues.


joelalejandro:
I'm merging this, everything looks correct on my end. :+1:

PR #229, opened 2020-02-23
Fix Included resources computed properties

spersico:
**Fixes all included resources not getting their computed properties computed.**

For example: on a _GET_ to both `/votes/1?include=article` and  `/votes/1?include=article.author`, where both **article** and **author** had computed properties, the computed properties were not being... well, computed.

**This PR fixes that. Both on direct and nested includes.**
Because before to this PR, the _"not computing on nested resources"_ was expected behaviour, the tests **must** be updated.

## UPDATE
As of 12/10/2020, this PR will be rebased on top of another that added testing for included resources, in order to both fix the test that need updating, and sort of putting a "readyness check" for included resources.

The tests involve having working GET operations as following:

- [ ] Basic GET /resourceA/id?include=resourceB.resourceC where resourceA *->resourceB *-> resourceC
- [ ] Basic GET /resourceA/id?include=resourceB.resourceC where resourceA *->resourceB 1->resourceC
- [ ] Basic GET /resourceB/id?include=resourceA.resourceC where resourceB 1->resourceA *-> resourceC
- [ ] Basic GET /resourceB/id?include=resourceA.resourceC where resourceB 1->resourceA 1-> resourceC


src/processors/operation-processor.ts
```diff
@@ -36,11 +36,54 @@ export default class OperationProcessor<ResourceT extends Resource> {
 
     if (result !== undefined) {
       eagerLoadedData = await this.eagerLoad(op, result);
+      eagerLoadedData = await this.computeRelationshipProperties(op, eagerLoadedData);
     }
 
     return this.convertToResources(op, result, eagerLoadedData);
   }
 
+  async computeRelationshipProperties(op, eagerLoadedData) {
```
joelalejandro, line 45: I'm thinking we should probably break down this function into smaller pieces. Take a look at the following comments.

src/processors/operation-processor.ts
```diff
@@ -36,11 +36,54 @@ export default class OperationProcessor<ResourceT extends Resource> {
...
+    const baseResourceClass = await this.resourceFor(op.ref.type);
+    for (const relationship in eagerLoadedData) {
+      if (eagerLoadedData.hasOwnProperty(relationship)) {
```
joelalejandro, line 48: This `if` is wrapping the whole thing. Let's skip the elements that do not meet this condition as early as possible. Also, we an use the `in` operator for making this statement a bit clearer:

```ts
// If the relationship isn't part of the loaded data,
// move on to the next relationship.
if (!(relationship in eagerLoadedData)) {
  continue;
}
```

src/processors/operation-processor.ts
```diff
@@ -36,11 +36,54 @@ export default class OperationProcessor<ResourceT extends Resource> {
...
+
+        const relationResourceClass = baseResourceClass.schema.relationships[relationship] && baseResourceClass.schema.relationships[relationship].type();
+        if (relationResourceClass) {
```
joelalejandro, line 51: Same here, a big `if` block. Let's negate it and exit as early as possible:

```ts
// If there's no class handling this type of resource, move on.
if (!relationResourceClass) {
  continue;
}
```

src/processors/operation-processor.ts
```diff
@@ -36,11 +36,54 @@ export default class OperationProcessor<ResourceT extends Resource> {
...
+          const includedRelationship = eagerLoadedData[relationship];
+          const resourceProcessor = await this.processorFor(relationResourceClass.type);
+
+          for (const includedRelationshipResource in includedRelationship["direct"]) {
```
joelalejandro, line 55: Can we type `includedRelationship` to access `direct` without the string literal? Also, to avoid repeating the expression all over this block, I'd rather store it in a variable:

```ts
const directRelations = includedRelationship.direct; // or ["direct"] if you must
```

joelalejandro: We could also probably have two lesser functions to handle each type of relationship and delegate these instructions. Something like `includeDirectRelationships()` and `includeNestedRelationships()`.

joelalejandro (changes requested):

spersico:
going to close this one, and make a new PR that handles both the computed resources and the modified testing on the whole, because handling testing of a fix and the fix itself in separate branches doesn't make any sense to me.

PR #233, opened 2020-10-12
Adds example for OperationProcessor

### LINKED ISSUE #214: OperationProcessors aren't accepted as type-valid processors in the Application configuration
Passing a processor that derives from `OperationProcessor` to the `Application` configuration results in the following error:

```
Type 'typeof PoemProcessor' is not assignable to type 'typeof OperationProcessor'.
  Type 'PoemProcessor' is not assignable to type 'OperationProcessor<ResourceT>'.
    Types of property 'execute' are incompatible.
      Type '(op: Operation) => Promise<void | Poem | Poem[]>' is not assignable to type '(op: Operation) => Promise<void | ResourceT | ResourceT[]>'.
        Type 'Promise<void | Poem | Poem[]>' is not assignable to type 'Promise<void | ResourceT | ResourceT[]>'.
          Type 'void | Poem | Poem[]' is not assignable to type 'void | ResourceT | ResourceT[]'.
            Type 'Poem' is not assignable to type 'void | ResourceT | ResourceT[]'.
              Type 'Poem' is not assignable to type 'ResourceT'.
                'Poem' is assignable to the constraint of type 'ResourceT', but 'ResourceT' could be instantiated with a different subtype of constraint 'Resource'.ts(2322)
```

A quick workaround is to force-cast it with the `as` operator:

```ts
const app = new Application({
  types: [Poem],
  processors: [PoemProcessor as typeof OperationProcessor]
});
```

joelalejandro:
Tested #214 and the error is no longer reproducing, so this PR adds an example of how to use an OperationProcessor-driven resource next to other KnexProcessor-based resources.

Resolves #214.

### DIFF
--- a/tests/dummy-app/processors/random.ts
+++ b/tests/dummy-app/processors/random.ts
@@ -1,0 +1,24 @@
+import Random from "../resources/random";
+import { OperationProcessor, HasId, Operation } from "../jsonapi-ts";
+import jsonApiErrors from "../../../src/errors/json-api-errors";
+
+const randomDataGenerator = {
+  number: () => ({ randomNumber: Math.random() }),
+  string: () => ({ randomString: parseInt(Math.random().toString().split(".")[1]).toString(16) }),
+  date: () => ({ randomDate: new Date(1602518929 + Math.random() * 10000000000000).toJSON() })
+}
+
+export default class RandomProcessor<ResourceT extends Random> extends OperationProcessor<ResourceT> {
+  static resourceClass = Random;
+
+  async get(op: Operation): Promise<HasId | HasId[]> {
+    if (op.ref.id in randomDataGenerator) {
+      return {
+        id: op.ref.id,
+        ...randomDataGenerator[op.ref.id]()
+      }
+    } else {
+      throw jsonApiErrors.BadRequest(`Allowed random data generators: ${Object.keys(randomDataGenerator).join(", ")}`);
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


tests/dummy-app/processors/random.ts
```diff
@@ -0,0 +1,24 @@
...
+const randomDataGenerator = {
+  number: () => ({ randomNumber: Math.random() }),
+  string: () => ({ randomString: parseInt(Math.random().toString().split(".")[1]).toString(16) }),
+  date: () => ({ randomDate: new Date(1602518929 + Math.random() * 10000000000000).toJSON() })
+}
+
+export default class RandomProcessor<ResourceT extends Random> extends OperationProcessor<ResourceT> {
+  static resourceClass = Random;
+
+  async get(op: Operation): Promise<HasId | HasId[]> {
```
spersico, line 14: this `| HasId[]` is necessary or correct?, the implementation of the function below seems to only use one id, unless I'm reading this wrong...

spersico (approved):
I quite like this one!
Though it would be nice to have a test or two added for this new example. Looks fine to me!

tests/dummy-app/processors/random.ts
joelalejandro: That's the signature for the `get` operation in the `OperationProcessor`:

https://github.com/kurierjs/kurier/blob/8c3f134dccdb1c868c545e9a054f1a97a13b7f1e/src/processors/operation-processor.ts#L123

PR #234, opened 2020-10-12
Adds NoOpTransaction for Knexless operations

### LINKED ISSUE #215: OperationProcessor-handled resources can't execute operations
Since #5 has not been implemented yet, `Application` requires a service that enables it to create transactions. The only service supported by JSONAPI-TS for this purpose is `knex`.

OperationProcessors do not require any external dependencies to work, but when an `Application` attempts to create a transaction in a "Knexless" environment and if the operation fails, the API crashes while attempting to use an unavailable `transaction.rollback()`:

```
    TypeError: Cannot read property 'rollback' of undefined"
```

A possible solution until a better solution is implemented is to provide a no-op transaction on `createTransaction()` when there is no `knex` service:

```ts
    if (!knex) {
      return {
        commit: () => {},
        rollback: () => {}
      };
    }
```

joelalejandro:
Resolves #215.

This PR implements the solution described in the issue, by creating a `NoOpTransaction` type and adding it as a possible type wherever `transaction` is instantiated.

### DIFF
--- a/src/application-instance.ts
+++ b/src/application-instance.ts
@@ -4,14 +4,14 @@ import * as Knex from "knex";
 import Application from "./application";
 import Resource from "./resource";
 import OperationProcessor from "./processors/operation-processor";
-import { Operation, OperationResponse } from "./types";
+import { Operation, OperationResponse, NoOpTransaction } from "./types";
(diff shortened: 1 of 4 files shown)


spersico (approved):
LGTM

PR #242, opened 2020-10-25
Add support for Node 14, Update most of the dependencies

spersico:
_...Some men just want to watch the world burn_

This PR updates most of the dependencies to the latest version where the update didn't introduce errors.

**The most important is sqlite3, which was blocking support for node 14, and typescript, which was really old.**

But all the tests pass, it builds perfectly fine, and I tested some of the test app endpoints in postman and everything looks fine.
image

There are 3 packages that introduced issues when trying to update to the latest version:
- **knex**, all versions after the current one (0.20.10) make some tests to stop passing. This is a problem that we'll need to solve at some point.
 - **typescript** throws an error after trying to update to the 4.0.X versions, some extra strict typing problem that I didn't want to deal with, jumping from 3.6.3 to 3.9.7 seemed good enough for now.
 - the **@types/express** types package was a clusterfuck of problems, I spent like half an hour trying to make it work, but it seems like the guys that maintained that package broke something in the minor/patch version. Honestly, I don't care enough about the Express types to deal with it, and right now it ain't broke, so meh 🤷🏻‍♂️.


joelalejandro (approved):
I'll add some issues to track Knex, TypeScript and @types/express.

PR #262, opened 2021-02-09
Refactors all package code to strictNullChecks = true

### LINKED ISSUE #250: Refactor code to be compliant with strictNullChecks enabled
We have over +40 non-null-safe issues in the library. Turning on `strictNullChecks` on `tsconfig` and refactoring as needed would greatly increase the project's robustness.

joelalejandro:
Resolves #250.

### DIFF
--- a/src/serializers/serializer.ts
+++ b/src/serializers/serializer.ts
@@ -45,10 +45,10 @@ export default class JsonApiSerializer implements IJsonApiSerializer {
     const primaryKey = resourceType.schema.primaryKeyName || DEFAULT_PRIMARY_KEY;
     const schemaRelationships = resourceType.schema.relationships;
     op.data.attributes = Object.keys(schemaRelationships)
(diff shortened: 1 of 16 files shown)


src/serializers/serializer.ts
```diff
@@ -213,12 +213,13 @@ export default class JsonApiSerializer implements IJsonApiSerializer {
                             attribute => subResourceClass.schema.attributes[attribute] === Password
                           )
                         ]),
-                        relationships: nestedResources.filter
+                        // A drunk Santiago walks in the bar...
+                        relationships: {} // nestedResources.filter
```
spersico, line 217: there's a story there, I'm sure...

If I could just remember 😝...

spersico (approved):
Awesome job! There were a couple of cases that I see that you type the same thing twice. I'm guessing that the type checker was being dumb, and yelling at you.

LGTM!

PR #271, opened 2021-03-12
Chore/update dependencies (Damn you Knex)

spersico:
So... yeah, my OCD struck again, and I started updating dependencies.
All of them were trivially easy to update.... 'till I got to Knex:

Seems like they [updated a bunch of times lately](https://github.com/knex/knex/blob/master/CHANGELOG.md#0952---11-march-2021), and we are lagging behind. It even has ESM and typescript support now! (too much support, they require typescript 4.1, which forced me to update that too).

They got a nifty [Update guide](https://github.com/knex/knex/blob/master/UPGRADING.md#upgrading-to-version-0950) which explains a bunch of things.

The bad thing is that some tests (6 out of 83) are failing, and I got stuck, and it's really late, so I'm going to sleep. I'll check it out again in the weekend, and if I don't get it then, I'll resign myself to a quieter life, and migrate to the mountains, to raise goats and die alone, all because of a few unhandled errors.


spersico:
hey @joelalejandro let me know when you can help me with this one. Been playing around, and I don't seem to get what is going on.

joelalejandro:
This work was completed in #279.

Issue #243, opened 2020-10-25
Resource Links support


spersico:
We decided to launch to V1 without this feature, but the plan is to someday add support for resource links, and related resource links.

On a first glance, we have 4 big features to implement: 
-  Provide Links of the the main resource ([ref](https://jsonapi.org/format/#document-resource-object-links))
-  Provide Links of the related resources ([ref](https://jsonapi.org/format/#document-resource-object-related-resource-links))
-  Provide GET access to the related resources on those links ([ref](https://jsonapi.org/format/#fetching-relationships))
-  Provide UPDATE access to the related resources on those links ([ref](https://jsonapi.org/format/#crud-updating-relationships))

isBatak:
Any updates on this one?
And what about pagination, should this be added to?

- `first`: the first page of data
- `last`: the last page of data
-` prev`: the previous page of data
- `next`: the next page of data

https://jsonapi.org/format/#fetching-pagination

isBatak:
Would it be possible to add pagination with a custom Addon?

spersico:
hi @isBatak!, it should... but the one that built the addons system is @joelalejandro.

@joelalejandro what do you think?

isBatak:
@joelalejandro 👋
Can you point out what should I do?
For the first step, I just want to support `first`, `last`, `prev`, `next` links for resources response.
I guess I need to extend this part https://github.com/kurierjs/kurier/blob/master/src/application.ts#L228 with links object, but how should I do it through Addon?

isBatak:
I figure out how to override `buildOperationResponse` inside of an addon.
```
import {
	Addon,
	Application,
	ApplicationInstance,
	HasId,
	KnexProcessor,
	Operation,
	OperationResponse,
	Resource,
} from 'kurier';
import { PaginationOptions } from './types';

const defaults: PaginationOptions = {
	defaultPaginator: 'paged',
	defaultPageSize: 10,
	maximumPageSize: 20,
};

export class PaginationAddon extends Addon {
	constructor(public readonly app: Application, public readonly options?: PaginationOptions) {
		super(app);
		this.options = { ...defaults, ...options };
	}

	async install(): Promise<void> {
		const oldBuildOperationResponse = this.app.buildOperationResponse.bind(this.app);

		this.app.buildOperationResponse = async (
			data: Resource | Resource[] | void,
			appInstance: ApplicationInstance
		): Promise<OperationResponse> => {
[...]

PR #275, opened 2021-05-13
Links specification compliance

isBatak:
This is still WIP, as I need some time to write tests.
I'll make it as a draft so I can refer to it in issue #243.

TODO:
- [x] include links in resources (self) and relationships (self and related)
- [x] move `baseUrl` to `LinkBuilder` inside serializer
- [x] pagination links
- [x] disable links per resource
- [ ] add `omitResourceIdentifiers` option to relationship (renamed to `alwaysIncludeLinkageData`) WIP
- [ ] ability to disable links globally
- [ ] implement links routes
- [ ] support meta in links
- [ ] fix all TODOs
- [ ] fix and write more tests
- [ ] write docs


src/types.ts
```diff
@@ -46,15 +46,37 @@ export type ResourceRelationshipData = {
...
+export type Meta = Record<string, AttributeValue>;
+
+export type Link = string | {
+  href: string;
+  meta?: Meta;
+};
+
+export type Links = {
+  self?: Link;
+  related?: Link;
```
isBatak, line 58: I think we should lose it up a bit because, in theory, it could contain any property.
For example https://github.com/json-api/json-api/pull/1515/files#diff-a80e2578f8c89d0702e3677e4f40fd400711ffd6be44a637a5e0b61560d378feL779

joelalejandro: We could use a `Partial<Link>` perhaps to lose it up a bit.

tests/test-suite/test-app/http.ts
```diff
@@ -4,9 +4,25 @@ import * as express from "express";
 import { jsonApiKoa, jsonApiExpress } from "./kurier";
 
 const koaApp = new Koa();
+koaApp.use(async (ctx, next) => {
+  if (ctx.req.url === '/favicon.ico') {
+    ctx.res.statusCode = 404;
+    ctx.res.end();
+  } else {
+    await next();
+  }
```
joelalejandro, line 13: This is just me being picky, but I think we can probably create a compatible middleware for both apps:

```ts
const faviconMiddleware = async (req, res, next) => {
  // ...
};

koaApp.use(async (ctx, next) => faviconMiddleware(ctx.req, ctx.res, next));
expressApp.use(faviconMiddleware);
```

src/application-instance.ts
```diff
@@ -11,8 +11,15 @@ import User from "./resources/user";
...
+  public baseUrl?: URL | undefined;
 
-  constructor(public app: Application) { }
+  constructor(public app: Application, baseUrl?: URL) {
+    if (!app.baseUrl && !baseUrl) {
+      throw Error('Application "baseUrl" is not defined!');
+    }
+
+    this.baseUrl = app.baseUrl || baseUrl;
+  }
```
joelalejandro, line 22: I'm not sure I understand the reason for having `baseUrl` available from either  `app.baseUrl` or the `baseUrl` parameter? Can you clarify for me?

src/types.ts
isBatak: That will only make them bot optional, which they already are. I was thinking to add some generic so we can extend it from the outside. 

Or something like this
 ` [Key: string]: T;`

tests/test-suite/test-app/http.ts
isBatak: hehe this is a funny story. While debugging the test-app in the browser I noticed that breakpoint inside operations get triggered twice in a row, but resource is undefined in the second pass. It took me a while until I figure out it was automatic favicon.ico browser request. Such a waste of time :)
I like the suggestion... will do it.

src/application-instance.ts
isBatak: We can build baseUrl from `req.protocol` and `req.host` and pass it to app instance. But if you run the API behind some proxy with different domain you maybe want to hardcode it or pass it through env. Also, WebSocket server needs the hardcoded value too.

isBatak:
Current progress:
1. Top-level document links are working
2. Relationships links are working
3. We can compute resource meta in the processor for exposing some read-only attributes (in my opinion `createdAt`, `updatedAt` and similar should be exposed in meta so no one can update them manually... but maybe I'm wrong so I used `stats` as an example) 
update

isBatak:
@joelalejandro @spersico 
Why is `id` optional in the Resource definition?
https://github.com/kurierjs/kurier/blob/master/src/resource.ts#L14
 
I'm having a problem with this definition inside `application.serializeResources` method
optional_id
I'm expecting to have `id` on the resource, is there any edge-case where `id` is `undefined`?

joelalejandro:
@isBatak `id` can be undefined when you're `POST`ing a new resource to the API and you're using database-generated IDs.

tests/test-suite/test-data/camelCase/seeds/seed.ts
```diff
@@ -43,7 +45,37 @@ exports.seed = (knex) => {
...
+    {
+      tableName: 'books',
+      values: [...Array(100)].map(() => ({
+        id: faker.datatype.uuid(),
+        title: faker.lorem.lines(1),
+        datePublished: faker.date.past().toISOString(),
+        isbn: faker.datatype.number({ min: 1000000, max: 9999999 }),
+        author: faker.datatype.number({ min: 1, max: 3 })
+      }))
+    },
```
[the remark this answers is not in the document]
joelalejandro: You could probably do something like:

```js
const initialData = [
   // ...
   ...new Array(10).fill(0).map(_ => ({
      tableName: 'books',
      values: [...Array(100)].map(() => ({
        id: faker.datatype.uuid(),
        title: faker.lorem.lines(1),
        datePublished: faker.date.past().toISOString(),
        isbn: faker.datatype.number({ min: 1000000, max: 9999999 }),
        author: faker.datatype.number({ min: 1, max: 3 })
      }))     
   })
];

File: tests/test-suite/test-data/snake_case/seeds/seed.ts [hunk 5]
```diff
@@ -1,3 +1,5 @@
+let faker = require('faker');
```
joelalejandro, line 1: @isBatak This should probably be `const` right? Unless I'm missing a reassignment somewhere?

isBatak: It seems like `requires` are shared between `snake_case` and `camelCase` seeds, so this was a workaround.
You can see in `camelCase` seeds I'm requiring it like this `faker = require('faker');`, basically reassigning the variable, that's why it's `let`. 
It's not ideal but it works. Also, it would be cool if we can use `import faker from 'faker';` syntax but it is not working inside seed files for some reason.

tests/test-suite/test-data/camelCase/seeds/seed.ts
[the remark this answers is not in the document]
isBatak: 🤦  I feel so stupid right now xD
But is there a way to increase the insert limit, is it Knex or SQLite thing? (I guess SQLite)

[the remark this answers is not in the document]
joelalejandro: @isBatak I just looked up, Knex offers a `batchInsert` utility for this purpose: https://knexjs.org/#Utility-BatchInsert

tests/test-suite/test-data/snake_case/seeds/seed.ts
joelalejandro: @spersico Any ideas here? I'm guessing it has something to do with the way we load the seed files?

joelalejandro:
Hey @isBatak, just a heads-up: #279 is being merged soon, and it'll upgrade all dependencies in the project. It might bring some conflicts because Knex changed the way its exports work, so you can expect to see a lot of stuff like this:

```diff
- import * as Knex from "knex";
+ import { Knex, knex } from "knex";
```

where uppercase-Knex is the namespace and lowercase-knex is the knex instance.

src/processors/knex-processor.ts
```diff
@@ -164,7 +174,14 @@ export default class KnexProcessor<ResourceT extends Resource> extends Operation
...
     }
-    return records;
+
+    return new ResourcesOperationResult(
+      HttpStatusCode.OK,
+      records,
+      {
+        recordCount,
+      }
+    );
```
joelalejandro, line 184: @isBatak Processors are by definition transport-layer-agnostic. Unfortunately, this means that we cannot have a `HttpStatusCode.OK` here. Maybe if you can tell me a bit more about your reasoning here, we can come up with a different approach.

isBatak: I kinda agree with that rule, but how will `transport-layer` know which status to return back to the client. 
That information is owned by Processor.
We need some kind of abstraction between Processor and Transport Layer because there is multiple status codes that could be retuned by the Processor based on the settings, like on create https://jsonapi.org/format/1.1/#crud-creating-responses. 
But yeah Maybe I'm missing some detail here.
Main idea of `ResourcesOperationResult` is that it solves pagination problem, it holds the subset (one page) of data plus pagination metadata used for building links (LinkBuilder). Maybe I can focus on that for now...

joelalejandro: Kurier not only works with HTTP transports, but also with WebSockets. Because of this, processors can't use HTTP status codes: they aren't part of the response for WebSockets. Unless something goes wrong, jsonApiKoa/jsonApiExpress/etc. return 200, 201 or 204 (according to the HTTP method). If something goes wrong in a processor, we throw exceptions from the `JsonApiErrors` object. This pretty much solves status codes for us.l 

AFAIK, the spec doesn't require us to return data about the current page, the record count and other pagination info, only the links. In any case, such information could be returned in the `meta` section of the document.

isBatak: > Kurier not only works with HTTP transports, but also with WebSockets. Because of this, processors can't use HTTP status codes: they aren't part of the response for WebSockets. Unless something goes wrong, jsonApiKoa/jsonApiExpress/etc. return 200, 201 or 204 (according to the HTTP method). If something goes wrong in a processor, we throw exceptions from the JsonApiErrors object. This pretty much solves status codes for us.  

I understand, and I agree... I'll remove `HttpStatusCode.OK` from the `ResourcesOperationResult`.

> AFAIK, the spec doesn't require us to return data about the current page, the record count and other pagination info, only the links. In any case, such information could be returned in the meta section of the document. Introducing the `ResourcesOperationResult` type might have a significant impact across the codebase.

Idea and problem that `ResourcesOperationResult` solves is this, when we hit endpoint like `/book` operation result is `Array<Record>` so I don't have a place where I can store pagination info, except modifying Array props, which is weird. 
Also, it could store JsonApiParams so I don't have to prop drill them through the application methods (buildOperationResponse, serializeResources), and serializer to be able to build links.

joelalejandro: > Idea and problem that `ResourcesOperationResult` solves is this, when we hit endpoint like `/book` operation result is `Array<Record>` so I don't have a place where I can store pagination info, except modifying Array props, which is weird. 

OK, so this type would be a counterpart to `Operation`. So in a request/response model, we'd have `Operation` for the request` and `OperationResult` for the response. In `OperationResult` we'd need something to store something like `data` and `metadata` (`data` being the resources, `metadata` being anything extra, such as pagination info). Am I getting this right? If that's the case, I believe we can go ahead.

> Also, it could store JsonApiParams so I don't have to prop drill them through the application methods (buildOperationResponse, serializeResources), and serializer to be able to build links.

`JsonApiParams` is always available in the `Operation` instance. I'm guessing you _could_ store that as part of the response metadata, if that makes things easier to work with, but I'm not entirely sure if it's the best way to do it.

isBatak:
@joelalejandro now when we have links "almost" finished should we consider removing linkages on relationships if URL doesn't include specific relationship.
For example, `article` has many `comments`
```
GET /article/1

{
  "data": {
    "id": "1",
    "type": "article",
    "attributes": {
      "body": ""
    },
    "relationships": {
      "comments": {
        "links": {
          "self": "http://localhost:3000/api/article/1/relationships/comments",
          "related": "http://localhost:3000/api/article/1/comments"
        },
        "data": [
          {
            "id": 1,
            "type": "comment"
          },
          {
            "id": 2,
            "type": "comment"
          },
          {
            "id": 3,
            "type": "comment"
          },
          {
            "id": 4,
            "type": "comment"
          },
          {
            "id": 5,
            "type": "comment"
          }
          // ..
        ]
      }
    },
    "links": {
      "self": "http://localhost:3000/api/article/1"
    }
  }
}

```
Should we somehow omit relationships `data` [resource identifiers](https://jsonapi.org/format/1.1/#document-resource-identifier-objects) for both `toOne` or `toMany` relationships if compound document isn't requested (request URL doesn't include).
This will potentially speed up the response because operation doesn't have to get the `comments` from DB to build linkage.

To get resource identifiers later on, we can fetch relationship `self` link and make a connection (this should be handled by client lib)
[From docs](https://jsonapi.org/format/1.1/#document-resource-object-relationships)
> When fetched successfully, this link returns the linkage for the related resources as its primary data. (See Fetching Relationships.)

To get actual resource data we can use `related` link.
The point is that we can cache those links and make response even faster. 

I'm thinking that we can opt-in to this behaviour in Resource relationships settings similar to this plugin https://github.com/kurierjs/kurier-addon-auto-include 
Something like:
```
import { Resource } from "kurier";
import Comment from "./comment";

export default class Article extends Resource {
  static schema = {
    attributes: {
      body: String,
    },
    relationships: {
      comments: {
        type: () => Comment,
        foreignKeyName: "comment_id",
        hasMany: true,
        omitResourceIdentifiers: true, // this could be false by default
      },
    },
  };
}
```

And in this case response will be:
```
GET /article/1

{
  "data": {
    "id": "1",
    "type": "article",
    "attributes": {
      "body": ""
    },
    "relationships": {
      "comments": {
        "links": {
          "self": "http://localhost:3000/api/article/1/relationships/comments",
          "related": "http://localhost:3000/api/article/1/comments"
        },
      }
    },
    "links": {
      "self": "http://localhost:3000/api/article/1"
    }
  }
}
```

and if we request Compound Document

```
GET /article/1?include=comments

{
  "data": {
    "id": "1",
    "type": "article",
    "attributes": {
      "body": ""
    },
    "relationships": {
      "comments": {
        "links": {
          "self": "http://localhost:3000/api/article/1/relationships/comments",
          "related": "http://localhost:3000/api/article/1/comments"
        },
        "data": [
          {
            "id": 1,
            "type": "comment"
          }
        ]
      }
    },
    "included": [{
      "type": "comment",
      "id": "1",
      "attributes": { ... }.
      "links": {
          "self": "http://localhost:3000/api/comment/1"
       },
    } ],
    "links": {
      "self": "http://localhost:3000/api/article/1"
    }
  }
}
```

joelalejandro:
Sounds great @isBatak, go ahead!

src/processors/knex-processor.ts
```diff
@@ -132,7 +132,9 @@ export default class KnexProcessor<ResourceT extends Resource> extends Operation
     const { type, schema } = this.resourceClass;
     const { attributes, relationships, primaryKeyName } = schema;
     const relationshipsKeys = Object.entries(relationships)
-      .filter(([, value]) => value.belongsTo)
+      .filter(([, value]) => {
+        return value.belongsTo && value.alwaysIncludeLinkageData;
```
isBatak, line 136: @joelalejandro I hope this is the right place to exclude linkages.
I have to figure how to handle few cases.
1. If `alwaysIncludeLinkageData` is false we should skip the relationship query 
2. if `alwaysIncludeLinkageData` is false and url contains `include=comments` then it should include linkage
3. if `alwaysIncludeLinkageData` is false and url contains `fields[articles]=comments`
4. if `alwaysIncludeLinkageData` is true then always include linkages

It should also take care of deep nested relationships, but that over my head right now.
Is there any other case? I need help with this one...

isBatak: Also, it seems that 2. case is the most difficult because I don't have any info about include.

joelalejandro: I think you might want to make such decisions in the `JsonApiSerializer` class. In particular, `serializeResource` and `serializeIncludedResources`.

isBatak: I'm pretty confident that it should be done here. 
It's the matter fo optimisation. Why should we build a complex DB query if we are not showing those resources in the response. It seems like a waste of processing time. 
Also, toMany resources could be filtered, paginated and sorted which also could be skipped if we relationships has links enabled.

joelalejandro: Alright, I guess you could add a third parameter that allows you to bring extra context to the scope of `getColumns`. In the `get()` method of the processor, you have `op`, which is the operation being resolved in the current request. There, you can check `params.include`. The [get()](https://kurier.readthedocs.io/en/latest/#get) docs show what information is available in this context.

isBatak: @joelalejandro I got stuck with the test cases for this example. I want to implement some real use-cases in test-app, for example ‘article’ has many ‘comments’ or something similar with ‘book’ resource, but I can’t make it work. Can you help me with this?

joelalejandro: @isBatak Sorry for the delay in my response! It's been a busy week. Can you tell me in greater detail what seems to be the problem? I'd be happy to help.

isBatak: @joelalejandro I'm trying to get response like this compound document https://jsonapi.org/format/1.1/#document-compound-documents but I can't find any example in the docs how to make `article` hasMany `comments` and get this part in the response:
```
"relationships": {
      "comments": {
        "data": [
          { "type": "comments", "id": "5" },
          { "type": "comments", "id": "12" }
        ]
      }
    }
 ```
 I would like to have this example in the `test-app`.

joelalejandro: I think the request should be `/article/1?include=comments` - @spersico can you help me out here?

isBatak: Idk why but comments are currently exposed on User resource... so this will not work. 
I guess I need to redesign the DB a bit ...

joelalejandro:
Hey @isBatak, just wanted to let you know you're doing a great job here <3 Sorry for the little-to-no-support these past few weeks, I've been busy getting stuff done at @somoscodear for our conference, <a href="https://webconf.tech">WebConf LATAM 2021</a>. Hopefully things will get calmer on September.

Keep up the great, great work!

joelalejandro:
Putting this PR on hold, since it hasn't had any activity in the past 6 months.

PR #284, opened 2021-06-21
Adds support for optional relationships in schema definitions

### LINKED ISSUE #232: Relationships shouldn't be required in the schema definition
If an entity has no relationships, we shouldn't be forcing the developer to explicitly say so.
Shouldn't the absence of `relationships` be enough?

For example, a `Tag` resource definition like this won't compile as of a8940048682ae36e1a57768ad4ed20dbbe813fd8:

```ts
class Tag extends Resource {
  static schema = {
    attributes: {
      name: String;
    }
  }
}
```

A very simple key-value resource is forced to have an explicit `relationships: {}` in its definition.

joelalejandro:
Resolves #232.

Some basic resources, such as classificators (i.e. resources with nothing more than an ID and a name), don't need to have relationships specified.

This PR modifies `resourceFor` to insert an empty (`{}`) object for `resource.schema.relationships` in case the definition doesn't include it.

This change has no impact on the `ResourceSchema` definition, since `relationships` will continue to be mandatory (otherwise, we'd have lots of `| undefined` casts to resolve, and that seemed like a lot of changes for a minor DX improvement).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,8 @@ A TypeScript framework to create APIs following the [1.1 Spec of JSONAPI](https:
      static schema = {
        attributes: {
          firstName: String,
-         lastName: String
+         lastName: String,
        },
-       relationships: {}
      };
    }
    ```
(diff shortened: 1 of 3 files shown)


spersico (approved):
LGTM

PR #285, opened 2021-06-21
Implements `jsonApiVercel` for Serverless Functions in Vercel

### LINKED ISSUE #282: Add support for Vercel serverless functions
Since Vercel doesn't allow to deploy custom Koa/Express servers, we should write an adapter that is compatible with the API framework Vercel uses.

joelalejandro:
Resolves #282.

Now that we've updated dependencies and released v1.1.0, we can move on with feature development. This PR introduces `jsonApiVercel`, a middleware for exposing a Kurier API as a serverless function.


joelalejandro:
@spersico The only issue with this is that I'm not sure of a good way of testing it. I was thinking about a script that runs a dummy nextjs app and then another one that runs `vercel dev` to test the Vercel-only use case.

Thing is - we'd need to install both as devDependencies, and *maybe* it feels like too much.

src/middlewares/json-api-vercel.ts
```diff
@@ -0,0 +1,45 @@
...
+    const appInstance = new ApplicationInstance(app);
+
+    try {
+      await authenticate(appInstance, req);
+    } catch (error) {
+      res.status(error.status);
+      res.json(convertErrorToHttpResponse(error));
+    }
+
+    if (process.env.AWS_REGION) {
```
spersico, line 25: This should IMO be added into the readme. This is a behavioural change that depends on the context of the app and could catch more than one dev unaware.

spersico (approved):

spersico:
> @spersico The only issue with this is that I'm not sure of a good way of testing it. I was thinking about a script that runs a dummy nextjs app and then another one that runs `vercel dev` to test the Vercel-only use case.
> 
> Thing is - we'd need to install both as devDependencies, and _maybe_ it feels like too much.

well.... as long as they are devDependencies, I don't think that it's THAT much. Maybe it could be set to run only on CI?

docs/index.md
```diff
@@ -641,6 +643,52 @@ The middleware, if successful, will respond with a `204 No Content` HTTP status
...
+    |__ [...kurier].js
+```
+
+In case you don't want all of your endpoints to go through Kurier, you can namespace them:
+
+```
+pages/
+|__ api/
+    |__ kurier/
+        |__ [...kurier.js]
```
isBatak, line 666:
```suggestion
        |__ [...kurier].js
```

docs/index.md
```diff
@@ -641,6 +643,52 @@ The middleware, if successful, will respond with a `204 No Content` HTTP status
...
+```
+
+In the `[...kurier.js]` file, you can import the `jsonApiVercel` middleware to handle the resource endpoints automatically, just like `jsonApiKoa` and `jsonApiExpress` do:
```
isBatak, line 669:
```suggestion
In the `[...kurier].js` file, you can import the `jsonApiVercel` middleware to handle the resource endpoints automatically, just like `jsonApiKoa` and `jsonApiExpress` do:
```

src/middlewares/json-api-vercel.ts
```diff
@@ -0,0 +1,45 @@
...
+      // This means we're not in localhost.
+      req["href"] = `https://${req.headers.host}${req.url!}`;
+    } else {
+      // This means we're in development. We won't use HTTPS here.
+      req["href"] = `http://localhost${req.url!}`;
+    }
+
+    req["urlData"] = urlData(appInstance, req.url!);
```
isBatak, line 33: It seems like `req.body` is a `string` on POST requests. I think it's because of nonstandard `
content-type:'application/vnd.api+json'` and Next.js don't know how to parse it.

isBatak: I had to do this it make it work
```
export default (req: NextApiRequest, res: NextApiResponse) => {
	if (req.body) {
		req.body = JSON.parse(req.body);
	}

	jsonApiVercel(app)(req, res);
};
```

src/types.ts
```diff
@@ -206,3 +209,20 @@ export type NoOpTransaction = {
 export type TransportLayerOptions = {
   httpBodyPayload?: string;
 }
+
+export type VercelRequest<BodyType = JsonApiDocument> = IncomingMessage & {
```
isBatak, line 213: I think `VercelRequest` and `VercelResponse` types should be replaced with Next.js ones
```
import type { NextApiRequest, NextApiResponse } from 'next'

export default (req: NextApiRequest, res: NextApiResponse) => {
  res.status(200).json({ name: 'John Doe' })
}
```

joelalejandro: Interesting, I hadn't noticed they're actually. I'll probably leave this one for Vercel (and not make any mentions to Next.js) and create a `jsonApiNext` middleware instead, to avoid confusion.

src/middlewares/json-api-vercel.ts
joelalejandro: Good catch! I'll add a conversion for it on the `jsonApiNext` middleware.

docs/index.md
```diff
@@ -641,6 +643,52 @@ The middleware, if successful, will respond with a `204 No Content` HTTP status
...
+// You can also add a database connection with Knex.
+app.services.knex = knex({ /* Your DB configuration */ });
+
+// Export the middleware result so Next.js can handle Kurier endpoints.
+export default jsonApiVercel(app);
+```
+
+#### Using Vercel Functions
+
+For Vercel Functions, the procedure is the same as above. Instead of creating `/pages/api/[...kurier.js]`, create `/api/[...kurier.js]`.
```
joelalejandro, line 690:
```suggestion
For Vercel Functions, the procedure is the same as above. Instead of creating `/pages/api/[...kurier].js`, create `/api/[...kurier].js`.
```

src/types.ts
isBatak: I get it now, this is just for Vercel functions and not for Next. Ok, makes sense…

src/middlewares/json-api-vercel.ts
isBatak: I just realized that this is because my client lib sends jsonapi Content-Type on every request. 
I have to check what happens if I don’t send Content-Type.

isBatak: If I don't send a content-type Next detect body as `content-type:'text/plain;charset=UTF-8'` and body is a string.
If I set `'content-type': 'application/json'`  then `body` is automatically parsed to object by Next.js.
Finally, if I set `'content-type': 'application/vnd.api+json'` body will be string again.

Should Kurier only allow `'content-type': 'application/vnd.api+json'`?
https://jsonapi.org/format/1.1/#introduction
> JSON:API requires use of the JSON:API media type (application/vnd.api+json) for exchanging data.

src/types.ts
joelalejandro: At first I thought this could be used for both, but it's better not get them mixed up.

joelalejandro:
I'll make changes in the docs to remove the mentiones to Next.js. That'll go in a separate middleware.

src/middlewares/json-api-vercel.ts
joelalejandro: While I agree that the required type is `vnd.api+json`, we wanted to make adoption easier and not change a thing in the frontend side. That is -- if you're already using `application/json`, we should be able to work with that.

isBatak: Maybe we could add `strict` mode and throw an error if `content-type` isn't correct...

joelalejandro: I like it. Such setting would live in the `Application` options, I imagine.

isBatak:
I just want to share my experimentation with this API.
Here is the repo https://github.com/isBatak/next-js-json-api-starter
It's live here https://next-js-kurier-starter.vercel.app/
I used https://app.supabase.io/ free plan to get Postgres DB, it was a pretty straightforward solution.

joelalejandro:
Excellent @isBatak! I discovered some issues with `?include=` queries and I'm working on fixing those.

joelalejandro:
> > @spersico The only issue with this is that I'm not sure of a good way of testing it. I was thinking about a script that runs a dummy nextjs app and then another one that runs `vercel dev` to test the Vercel-only use case.
> > Thing is - we'd need to install both as devDependencies, and _maybe_ it feels like too much.
> 
> well.... as long as they are devDependencies, I don't think that it's THAT much. Maybe it could be set to run only on CI?

Hmm, if we're putting them in `devDependencies`, I'd rather run the tests locally as well. Making this available just in CI feels like a DX friction.

src/middlewares/json-api-vercel.ts
joelalejandro: I ended up removing this, it's unnecessary.

joelalejandro:
@isBatak Latest changes to `jsonApiVercel` are available for usage in `kurier@1.2.0-alpha2`, in case you want to try them out!

PR #292, opened 2021-06-24
Adds `strict` mode for HTTP transport layers

### LINKED ISSUE #290: Add `strict` mode to require `application/vnd.api+json` in `Content-Type` header
If I don't send a content-type Next detect body as `content-type:'text/plain;charset=UTF-8'` and body is a string.
If I set `'content-type': 'application/json'`  then `body` is automatically parsed to object by Next.js.
Finally, if I set `'content-type': 'application/vnd.api+json'` body will be string again.

Should Kurier only allow `'content-type': 'application/vnd.api+json'`?
https://jsonapi.org/format/1.1/#introduction
> JSON:API requires use of the JSON:API media type (application/vnd.api+json) for exchanging data.

_Originally posted by @isBatak in https://github.com/kurierjs/kurier/pull/285#discussion_r656806119_

joelalejandro:
Resolves #290.

As described in the issue, the JSONAPI spec requires `application/vnd.api+json` as `Content-Type` for all requests. This PR adds support for opt-in usage of the spec's header.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1291,7 +1291,11 @@ import Author from "./resources/author";
 const app = new Application({
   namespace: "api",
   types: [Author],
-  defaultProcessor: new KnexProcessor(/* knex options */)
+  defaultProcessor: new KnexProcessor(/* knex options */),
+  transportLayerOptions: {
+    httpBodyPayload: '1mb',
+    httpStrictMode: true
+  },
 });
 
 const api = new Koa();
@@ -1307,6 +1311,9 @@ The `Application` object is instantiated with the following settings:
 - `types`: A list of all resource types declared and handled by this app.
 - `processors`: If you define custom processors, they have to be registered here as instances.
 - `defaultProcessor`: All non-bound-to-processor resources will be handled by this processor.
+- `transportLayerOptions`: Allows to customize some aspects of the transport layer used by the application.
+  - `httpBodyPayload`: Describes how big the request's `body` can be, expressed in a string i.e. `10mb`. Defaults to `1mb`.
+  - `httpStrictMode`: If enabled, requires all HTTP incoming requests to use `Content-Type: application/vnd.api+json`. Defaults to `false`.
 
 ### Referencing types and processors
 
--- a/src/middlewares/json-api-koa.ts
+++ b/src/middlewares/json-api-koa.ts
@@ -10,8 +10,22 @@ import {
   handleJsonApiEndpoint,
   urlData
 } from "../utils/http-utils";
+import jsonApiErrors from "../errors/json-api-errors";
 
 export default function jsonApiKoa(app: Application, ...middlewares: Middleware[]) {
+  const checkStrictMode = async (ctx: Context, next: () => Promise<any>) => {
+    if (!app.transportLayerOptions.strictMode) {
+      return next();
+    }
+
+    if (ctx.headers["content-type"] !== 'application/vnd.api+json') {
+      ctx.status = 400;
+      ctx.body = convertErrorToHttpResponse(jsonApiErrors.BadRequest("Content-Type must be application/vnd.api+json"));
+    }
+
+    return next();
+  };
+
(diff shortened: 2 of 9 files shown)


joelalejandro:
This feature is available in `kurier@1.2.0-alpha3`.

docs/index.md
```diff
@@ -1307,6 +1311,9 @@ The `Application` object is instantiated with the following settings:
 - `types`: A list of all resource types declared and handled by this app.
 - `processors`: If you define custom processors, they have to be registered here as instances.
 - `defaultProcessor`: All non-bound-to-processor resources will be handled by this processor.
+- `transportLayerOptions`: Allows to customize some aspects of the transport layer used by the application.
+  - `httpBodyPayload`: Describes how big the request's `body` can be, expressed in a string i.e. `10mb`. Defaults to `1mb`.
+  - `httpStrictMode`: If enabled, requires all HTTP incoming requests to use `Content-Type: application/vnd.api+json`. Defaults to `false`.
```
isBatak, line 1316: I had an idea of global `strictMode: true` that could be used for multiple things, not just for content-type.
For example, profiles should also be included in content-type 
`Content-Type: application/vnd.api+json; profile="https://example.com/resource-timestamps"`
https://jsonapi.org/format/1.1/#profile-rules

Extensions also:
`Content-Type: application/vnd.api+json; ext="https://jsonapi.org/ext/version"`
https://jsonapi.org/format/1.1/#extension-rules

We might find other loose rules in the future and throw an error in strict mode if they are not followed.

docs/index.md
```diff
@@ -1307,6 +1311,9 @@ The `Application` object is instantiated with the following settings:
 - `types`: A list of all resource types declared and handled by this app.
 - `processors`: If you define custom processors, they have to be registered here as instances.
 - `defaultProcessor`: All non-bound-to-processor resources will be handled by this processor.
+- `transportLayerOptions`: Allows to customize some aspects of the transport layer used by the application.
+  - `httpBodyPayload`: Describes how big the request's `body` can be, expressed in a string i.e. `10mb`. Defaults to `1mb`.
```
isBatak, line 1315: Isn't this a middleware concern?
For example, in Next.js it's defined inside api route file, a.k.a `api/[...kurier].js`
https://nextjs.org/docs/api-routes/api-middlewares#custom-config

docs/index.md
joelalejandro: For now, the only thing we check in `strictMode` is the Content-Type. Eventually we can keep adding more features to it. We do not support profiles or extensions at present, so we can make the necessary adjustments in that moment, I think.

docs/index.md
joelalejandro: It's been there since `1.0.0`. We wanted to keep the jsonApi* middlewares as simple as possible and `transportLayerOptions` seemed like a nice place back then.

Revisiting the idea, we might want to add support for also configuring `transportLayerOptions` directly into the middleware. To prevent breaking changes for now, we'd support both ways.

spersico: something to add to v2 maybe? We can break compatibility then 😉

src/middlewares/json-api-koa.ts
```diff
@@ -10,8 +10,22 @@ import {
   handleJsonApiEndpoint,
   urlData
 } from "../utils/http-utils";
+import jsonApiErrors from "../errors/json-api-errors";
 
 export default function jsonApiKoa(app: Application, ...middlewares: Middleware[]) {
+  const checkStrictMode = async (ctx: Context, next: () => Promise<any>) => {
```
spersico, line 16: is this... an `any`??? 😲

spersico (approved):

docs/index.md
joelalejandro: Sounds good

src/middlewares/json-api-koa.ts
joelalejandro: YASSSSSS.

PR #300, opened 2021-07-04
Prevents usage of invalid operators in filters

### LINKED ISSUE #297: Dynamic operator invocation could throw an exception
`operators[operator]` in line 149 could be `undefined`. 

From CodeQL: 

> Invocation of method with user-controlled name may dispatch to unexpected target and cause an exception.

https://github.com/kurierjs/kurier/blob/0b2e0156fb727fed89da90d6dfff9949bd6c9dbd/src/processors/operation-processor.ts#L145-L155

joelalejandro:
Resolves #297.


spersico (approved):
LGTM

PR #320, opened 2022-04-10
Dependency Updates September 2022

spersico:
Here we go again...

Updated every dependency except the latest "escape-string-regexp" (because of the mega-breaking change that it has).
This includes even internal deps _(I deleted the package-lock.json and ran npm install and it still worked!)_

And I got to this:
image😁

Beware, Knex 1.0.5 was a pain in the ass to get it to a working state (it seems like there are issues with node-sqlite3... again). So please test it on your Mac @joelalejandro, because the issues that I saw reported were related to Mac installs.

And no, this Knex version **still** lacks support for the _returning_ syntax 😭


joelalejandro:
I'll be sure to run all tests in a Mac OS environment this week :D

spersico:
Updated this to the latest version of every dependency, except for the following:
image

Basically, they break too many tests, that I don't feel like fixing ATM 😅

joelalejandro:
Will test tonight!

joelalejandro:
Confirmed in Mac OS that tests and installation work fine. Will attempt to update remaining dependencies.

joelalejandro:
Confirmed Mac OS installation & tests are OK again. Merging and releasing new version!

Issue #326, opened 2022-11-11
Add `meta` to operation responses, both at resource-level and document-level


joelalejandro:
Currently, Kurier can parse `meta` in requests, but it cannot inject `meta` in an operation response. We need a way to inject `meta` in resources and documents.

The proposal is to add the `meta` property at the `Resource` type and add hook-like functions to `OperationProcessor` so the `executeOperation()` method can call them if needed:

- Resource-level meta hooks:
  - `resourceMeta`: for any operation
  - `resourceMetaFor*`: for specific operations (get, add, update, custom ops)
- Document-level meta hooks:
  - `meta` instance property: for any operation
  - `metaFor*`: for specific operations (get, add, update, custom ops)

Initially we won't support `meta` on `remove` operations, because we are not returning any response on transport layers (i.e. HTTP 204 No Content).

Execution of these hooks must take place at the Application level, in the `executeOperation` method, prior to calling `buildOperationResponse`.

### Implementation example

```typescript
export default class ArticleProcessor<ResourceT extends Article> extends KnexProcessor<ResourceT> {
  /**
   * Returns a `meta` object at the processor (document) level for any operations.
   */
  async meta(resourceOrResources: Resource | Resource[]): Meta {
    // ...
  }

  /**
   * Returns a `meta` object at the resource level for any operations.
   */
  async resourceMeta(resource: Resource): Meta {
    // ...
  }

  /**
   * Returns a `meta` object at the processor (document) level, providing the operation
   * to help decide in which cases add a `meta`.
   */
  async metaFor(op: Operation, resourceOrResources: Resource | Resource[]): Meta {
    switch (op.op) {
      case 'get':
        // ...
      break;
      // ...
    }
    // ...
  }

  /**
   * Returns a `meta` object at the processor (document) level for `get` operations.
   */
  async metaForGet(resourceOrResources: Resource | Resource[]): Meta {
    // ...
  }

  /**
   * Returns a `meta` object at the processor (document) level for `add` operations.
   */
  async metaForAdd(resourceOrResources: Resource | Resource[]): Meta {
    // ...
  }

  /**
   * Returns a `meta` object at the processor (document) level for `update` operations.
   */
  async metaForUpdate(resourceOrResources: Resource | Resource[]): Meta {
    // ...
  }

  /**
   * Returns a `meta` object at the processor (document) level for a custom operation called `print`.
   */
  async metaForPrint(resourceOrResources: Resource | Resource[]): Meta {
    // ...
  }

[...]

PR #327, opened 2022-11-12
Support for `meta` in responses

joelalejandro:
This PR implements the `meta` hook API described in #326, allowing to add metadata objects both at resource-level and document-level, as described in the JSON:API spec:

- [Document meta](https://jsonapi.org/format/#document-meta)
- [Resource objects](https://jsonapi.org/format/#document-resource-objects)

This PR does not include support for `meta` in `links`, since #243 is still not available.

### DIFF
--- a/src/application.ts
+++ b/src/application.ts
@@ -16,8 +16,12 @@ import {
   OperationResponse,
   ResourceSchemaRelationship,
   NoOpTransaction,
+  MaybeMeta,
+  Meta,
 } from "./types";
 import flatten from "./utils/flatten";
+import { classify } from "./utils/string";
+import { isEmptyObject } from "./utils/object";
 
 export default class Application {
   namespace: string;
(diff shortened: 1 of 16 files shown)


joelalejandro:
Some tests breaking yet ☠️

src/application.ts
```diff
@@ -145,15 +151,81 @@ export default class Application {
...
+      }
+    }
+  }
+
+  async buildDocumentMeta(
+    data: Resource | Resource[] | void,
+    processor: OperationProcessor<Resource>,
+    op: Operation,
+  ): Promise<MaybeMeta> {
+    if (!data) {
```
marcemira, line 215: Can we please add the `OPERATIONS_INCOMPATIBLE_WITH_META_INJECTION` return here as well? We call `buildDocumentMeta` every time we want to `buildOperationResponse` (https://github.com/kurierjs/kurier/pull/327/files#diff-16a8f18a1e6d24ab6125db9b965edc8d7970d184e44ada35f3ccbf71a21670a4R320)

marcemira: Thanks a lot @joelalejandro 🎉 ✨

PR #367, opened 2023-10-28
fix: type of params.page.*

nelson6e65:
Those values are actually strings.

```
?page[limit]=19&page[number]=19&page[after:id]=SayLOC5U5pMj2KQlxCo0
```

```js


{page: { limit: '19', number: '19', 'after:id': 'SayLOC5U5pMj2KQlxCo0' }}
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -107,7 +107,7 @@ export type JsonApiParams = {
   include?: string[];
   sort?: string[];
   filter?: { [key: string]: string };
-  page?: { [key: string]: number };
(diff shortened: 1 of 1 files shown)


joelalejandro (approved):
Thanks @nelson6e65 !

joelalejandro:
@nelson6e65 Unfortunately I need to revert this because KnexProcessor uses this to do math around the paging. It's weird that the CI check didn't catch it :/

PR #370, opened 2023-11-07
fix(json-api-express): make sure response status is a number

### LINKED ISSUE #369: [1.3.x] errors.status should be a string
The [JSON:API specification](https://jsonapi.org/format/#error-objects) states:

> _"[...] the HTTP status code applicable to this problem, **expressed as a string value**"_
> 

But in the JSON response, it is a `number` instead of `string`.

```json
{
  "errors": [
    {
      "status": 501
    }
  ]
}
```

And if I use `string`, I get this warning:

> **Warning**
> express deprecated res.status("501"): use res.status(501) instead

😅

nelson6e65:
Regardless of Error instance has `status` property as string, convert it to number for `res.status()`.

Fixes #369

### DIFF
--- a/src/middlewares/json-api-express.ts
+++ b/src/middlewares/json-api-express.ts
@@ -51,7 +51,7 @@ export default function jsonApiExpress(
     try {
       await authenticate(appInstance, req);
     } catch (error) {
-      res.status(+error.status).json(convertErrorToHttpResponse(error));
+      res.status(Number(error.status)).json(convertErrorToHttpResponse(error));
       return next();
     }
 
--- a/src/middlewares/json-api-express.ts
+++ b/src/middlewares/json-api-express.ts
@@ -51,7 +51,7 @@ export default function jsonApiExpress(
     try {
       await authenticate(appInstance, req);
     } catch (error) {
-      res.status(Number(error.status)).json(convertErrorToHttpResponse(error));
+      res.status(+error.status).json(convertErrorToHttpResponse(error));
       return next();
     }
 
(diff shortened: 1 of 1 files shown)


src/middlewares/json-api-express.ts
```diff
@@ -51,7 +51,7 @@ export default function jsonApiExpress(
     try {
       await authenticate(appInstance, req);
     } catch (error) {
-      res.status(error.status).json(convertErrorToHttpResponse(error));
+      res.status(+error.status).json(convertErrorToHttpResponse(error));
```
joelalejandro, line 54: @nelson6e65 I agree with this, I'm going to apply it with `Number` though for a more explicit and readable code:

```suggestion
      res.status(Number(error.status)).json(convertErrorToHttpResponse(error));
```

joelalejandro (approved):

src/middlewares/json-api-express.ts
nelson6e65: You can use `Number.parseInt()` then, @joelalejandro

nelson6e65:
Ah, I remember that type problem, @joelalejandro . That's why I used `+`. You can use `Number.parseInt(`${error.status}`, 10)`

joelalejandro:
Nah, haha, this was a readibility change, I'd rather keep the + in that case. Reverting that last commit.

joelalejandro:
Hmm, somehow I'm unable to revert it, would you mind submitting this change again? Thanks and apologies <3

PR #373, opened 2023-12-22
fix/multiple.includes.nestedResources

Martinarbez:
This Pull Request introduces several changes to address a key issue and improve the functionality and testing of the application:

**Problem Solved**: The application was previously unable to handle queries for a resource with multiple nested resources simultaneously. If multiple nested resources were included in the query, only the last one was returned. This issue has been resolved with the changes in this PR.

1. **Enhancement in `knex-processor.ts`**: The `parseOperationIncludedRelationships` function has been updated to ensure that the `acumRelationships` object correctly accumulates relationships without overwriting existing ones. This is achieved by spreading the existing relationships before adding a new one. This change is part of the solution to the problem described above.

2. **Refactoring in `serializer.ts`**: The serialization process has been simplified. Previously, the code was creating a new array with unique `type_id` combinations from `includedData` and then mapping over this array to find the corresponding items in `includedData`. This has been replaced with a direct return of `includedData`, which simplifies the code and may improve performance.

3. **Addition of a new test in `nestedResources.test.ts`**: A new test has been added to verify the correct retrieval of the first article's author's votes and comments. This test will help ensure that the application's data retrieval and serialization processes are working as expected, and that multiple nested resources can be retrieved simultaneously.

### DIFF
--- a/src/processors/knex-processor.ts
+++ b/src/processors/knex-processor.ts
@@ -66,6 +66,7 @@ const parseOperationIncludedRelationships = (
     .filter((include) => include.length > 1)
     .reduce((acumRelationships, [nestedOrigin, nestedRelationshipName]) => {
       acumRelationships[nestedOrigin] = {
+        ...acumRelationships[nestedOrigin],
         [nestedRelationshipName]: relationships[nestedOrigin].type().schema.relationships[nestedRelationshipName],
       };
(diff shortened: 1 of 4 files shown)


joelalejandro (approved):
Excellent work!

